PR #92, opened 2021-04-07
Bug 1946273 - Replace busybox with UBI for VDDK image build

fdupont-redhat:
The Universal Base Image (UBI) provided by Red Hat is scanned for vulnerabilities, so using it ensures better security.


apinnick:
LGTM

PR #152, opened 2021-06-15
MTV-67: Filter must-gather data collection

apinnick:
https://issues.redhat.com/browse/MTV-67

Preview: https://deploy-preview-152--forklift-documentation.netlify.app/documentation/doc-migration_toolkit_for_virtualization/master/#using-must-gather_forklift

Ready for QE.

### DIFF
--- a/documentation/modules/using-must-gather.adoc
+++ b/documentation/modules/using-must-gather.adoc
@@ -5,23 +5,24 @@
 [id="using-must-gather_{context}"]
 = Using the must-gather tool
 
-You can collect logs, metrics, and information about {project-short} custom resources by using the `must-gather` tool.
+You can collect logs and information about {project-short} custom resources (CRs) by using the `must-gather` tool. You can gather data for a specific namespace, migration plan, or virtual machine (VM) by using the filtering options.
 
-The `must-gather` data must be attached to all customer cases.
+[NOTE]
+====
+You must attach a `must-gather` data file to all customer cases.
+====
 
-You can collect data for a one-hour or a 24-hour period and view the data with the Prometheus console.
+You can collect metrics data for a 24-hour period and view the data with the Prometheus console.
 
 .Prerequisites
 
 * You must be logged in to the {virt} cluster as a user with the `cluster-admin` role.
 * You must have the link:https://docs.openshift.com/container-platform/{ocp-version}/cli_reference/openshift_cli/getting-started-cli.html[{ocp} CLI (`oc`)] installed.
 
-.Procedure
+.Collecting logs and CR information
 
 . Navigate to the directory where you want to store the `must-gather` data.
 . Run the `oc adm must-gather` command:
-
-* To gather data for the past hour:
 +
 [source,terminal,subs="attributes+"]
 ----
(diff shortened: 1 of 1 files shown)


documentation/modules/using-must-gather.adoc
```diff
@@ -19,9 +21,7 @@ You can collect data for a one-hour or a 24-hour period and view the data with t
 .Procedure
 
 . Navigate to the directory where you want to store the `must-gather` data.
-. Run the `oc adm must-gather` command:
-
-* To gather data for the past hour:
+. Run the `oc adm must-gather` command to gather data for the past hour:
```
aufi, line 24: "Past hour" or 24 hours is not (and was not) part of Forklift must-gather, there is no such time constraint - please remove it.

documentation/modules/using-must-gather.adoc
```diff
@@ -30,11 +30,38 @@ $ oc adm must-gather --image={must-gather}
...
+
+* Virtual machine:
++
+[source,terminal,subs="attributes+"]
+----
+$ oc adm must-gather --image={must-gather} \
+  -- VM=<vm_id> NS=<namespace> /usr/bin/targeted
+----
+
+. Optional: Run the `oc adm must-gather` command with the `gather_metrics_dump` script to gather data for the past 24 hours:
```
aufi, line 59: Same as above, please remove "the past 24 hours".

documentation/modules/using-must-gather.adoc
```diff
@@ -30,11 +30,38 @@ $ oc adm must-gather --image={must-gather}
...
+$ oc adm must-gather --image={must-gather} \
+  -- PLAN=<migration_plan> /usr/bin/targeted
+----
+
+* Virtual machine:
++
+[source,terminal,subs="attributes+"]
+----
+$ oc adm must-gather --image={must-gather} \
+  -- VM=<vm_id> NS=<namespace> /usr/bin/targeted
```
aufi, line 56: Just for better understading what VM ID it is (not an ID from vmware/rhv/openshift/..), the VM ID is the one from the Plan CR

aufi (changes requested):

documentation/modules/using-must-gather.adoc
apinnick: The content in this section is based on the konveyor/forklift-must-gather. The [gather_metrics_dump](https://github.com/konveyor/forklift-must-gather/blob/main/collection-scripts/gather_metrics_dump) script says that it gathers data for 24 hours. Can you confirm that it does not?

documentation/modules/using-must-gather.adoc
apinnick: Is forklift must gather not using the scripts in konveyor? It looks like the [default](https://github.com/konveyor/forklift-must-gather/blob/main/collection-scripts/gather_metrics) is to collect data for 6 hours.

documentation/modules/using-must-gather.adoc
aufi: Ah, sorry, you're right the gather_metrics_dump script is for 24 hours, so it should be kept.   (just the "default" must gather execution doesn't have a time restriction, but also doesn't dump metrics).

documentation/modules/using-must-gather.adoc
aufi: Correct, the metrics capture script is there, but currently it is not executed by default by the main script https://github.com/konveyor/forklift-must-gather/blob/main/collection-scripts/gather

apinnick: Yes, I see what you mean. So it looks like the default gathers data about CRs and logs but NOT metrics data. Is that correct?

documentation/modules/using-must-gather.adoc
apinnick: Got it. metrics_dump collects metrics data for 24 hours but default gather only collects logs and CRs. Will fix.

documentation/modules/using-must-gather.adoc
aufi: Correct!

documentation/modules/using-must-gather.adoc
aufi: Correct, thank you!

aufi (approved):

PR #154, opened 2021-06-16
MTV-49: Minimal hooks

apinnick:
https://issues.redhat.com/browse/MTV-49

Add minimal hook documentation

Preview: https://deploy-preview-154--forklift-documentation.netlify.app/documentation/doc-migration_toolkit_for_virtualization/master/#creating-migration-plan_forklift

### DIFF
--- a/documentation/modules/migrating-virtual-machines-cli.adoc
+++ b/documentation/modules/migrating-virtual-machines-cli.adoc
@@ -12,6 +12,7 @@ You can migrate virtual machines (VMs) from the command line (CLI) by creating t
 * `Host` contains the VMware host details.
 * `NetworkMap` maps the source and destination networks.
 * `StorageMap` maps the source and destination storage.
+* Optional: `Hook` CR contains custom code that you can run on a VM at a specified phase of the migration.
 * `Plan` contains a list of VMs to migrate and specifies whether the migration is cold or warm. The `Plan` references the providers and maps.
 * `Migration` runs the `Plan`. If the migration is warm, it specifies the cutover time.
 +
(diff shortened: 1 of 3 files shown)


documentation/modules/migrating-virtual-machines-cli.adoc
```diff
@@ -222,6 +251,9 @@ EOF
 <5> Specify the name of the `StorageMap` CR.
 <6> You can use either the `id` _or_ the `name` parameter to specify the source VMs.
 <7> Specify the managed object reference of the VMware VM.
+<8> Optional: You can specify up to two hooks for a VM. Each hook must run during a separate migration step.
+<9> Specify the name of the `Hook` CR.
+<10> Allowed values are `PreHook`, before disk copy begins, or `PostHook`, after the image conversion.
```
apinnick, line 256: It's not clear to me what "image conversion" refers to.

apinnick:
@istein1 
Ready for QE.

documentation/modules/migrating-virtual-machines-cli.adoc
ibragins: When migration is going from VMWare, disk format is proprietary. When we are migrating to OCP, there are 2 actions:
a. Disk is copied
b. Disk format is converting from VMWare format to QCOW2, which is opensource. 

So, it means that hook can be performed as:
1. PreHook, actions will be taken before the plan will actually start performing
2. AfterHook, actions will be taken after that disk image conversion will be finished, which actually means that migration from VMWare is completed successfully.

Please take into account, that this is relevant ONLY for VMWare, for RHV no disk conversion is needed, just disk copying occurs.

PR #156, opened 2021-06-17
MTV-70: CBT workaround

apinnick:
https://issues.redhat.com/browse/MTV-70

Add CBT bug and workaround to Known Issues.

Preview: https://deploy-preview-156--forklift-documentation.netlify.app/documentation/doc-migration_toolkit_for_virtualization/master/#known-issues_forklift

### DIFF
--- a/documentation/modules/known-issues.adoc
+++ b/documentation/modules/known-issues.adoc
@@ -7,6 +7,22 @@
 
 This section describes known issues and mitigations.
 
-.Network map remains in a NotReady state
+.Network map displays a `Destination network not found` error
 
-If you are mapping more than one source and destination network and the network map remains in a `NotReady` state, the cause is a missing network attachment definition. You must create a link:https://docs.openshift.com/container-platform/{ocp-version}/virt/virtual_machines/vm_networking/virt-attaching-vm-multiple-networks.html#virt-creating-network-attachment-definition[network attachment definition] for each additional destination network before you create the network map. (*link:https://bugzilla.redhat.com/show_bug.cgi?id=1971259[BZ#1971259]*)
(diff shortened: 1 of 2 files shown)


mguetta1:
Looks good to me

documentation/modules/known-issues.adoc
```diff
@@ -7,6 +7,22 @@
...
+
+.Warm migration gets stuck during third precopy
+
+Warm migration uses changed block tracking snapshots to copy data during the precopy stage. The snapshots are created at one-hour intervals by default. When a snapshot is created, its contents are copied to the destination cluster. However, when the third snapshot is created, the first snapshot is deleted and the block tracking is lost. (link:https://bugzilla.redhat.com/show_bug.cgi?id=1969894[*BZ#1969894*])
+
+You can do one of the following to mitigate this issue:
+
+* Start the cutover stage no more than one hour after the precopy stage begins so that only one snapshot is created.
```
mguetta1, line 20: I think that it will be more accurate if we say:
Start the cutover stage no more than one hour after the precopy stage begins so that only **2 snapshots** are created, one in the internal prcopies stage and one in the cutover stage 
(cutover stage includes the final precopy, this is why the user has to set the cutover after one internal precopy)

mguetta1 (changes requested):

documentation/modules/known-issues.adoc
mguetta1: Or just adding "one internal snapshot is created"

apinnick: You are correct. Miguel wanted a more cautious wording. But I'm happy to change it as you suggest.

PR #185, opened 2021-08-04
MTV-80: Performance recommendations (kbase)

apinnick:
https://issues.redhat.com/browse/MTV-80

Turn [MTV 2.0 Scale & Performance](https://docs.google.com/document/d/11zXodYcykdtaUgjeAduutG4Ow--hqxje6hWUDZXBNlw/edit#) into a [Kbase article](https://access.redhat.com/articles/6242511) (not accessible because it is still a draft)

Preview: https://deploy-preview-185--forklift-documentation.netlify.app/documentation/performance/performance-recommendations/

### DIFF
--- a/documentation/performance/performance-recommendations.md
+++ b/documentation/performance/performance-recommendations.md
@@ -1,0 +1,109 @@
+The following recommendations can improve performance when migrating virtual machines (VMs) from VMware vSphere to OpenShift Virtualization using the Migration Toolkit for Virtualization (MTV).
+
+**These recommendations are not requirements for migration.**
+
+## Use fast, dedicated networks for the migration
+
+Use networks with a minimum transmission speed of 10 GbE in both the VMware and OpenShift Virtualization environments.
+
+Use dedicated networks to ensure that other services are not affected by the migration process.
+
+These recommendations also apply to a network selected as an [ESXi endpoint](#esxi-endpoint).
+
+## Select an appropriate endpoint
+
+The endpoint affects the ESXi data store read rate, which impacts the disk transfer speed and migration duration.
+
+The average read rate for the vSphere endpoint is 800 MBps, while the average read rate for the ESXi endpoint is 300-500 MBps. See [BZ#1957985](https://bugzilla.redhat.com/show_bug.cgi?id=1957985) for details.
+
+Select the endpoint that is appropriate for your vSphere version and number of hosts:
+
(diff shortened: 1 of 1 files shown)


mordechai:
@apinnick 
My suggestions are:
- '_Fast, dedicated networks for the migration_' should include line about datastore speeds, migrations are bound by esx datatstore speed, then network.
- _Maximum recommended number of VMs per host_ section requires rewording see below[1]
- _Maximum recommended number of VMs per host_ table should use index small 2 also here as done in previous table indicating that '2 Exceeding this value does not improve performance.'
- In _Maximum number of concurrent disk transfers_ / _Migration Example_  would be good to include esx datastore read rate.
- consider slight rewording of '_The following table displays the improved performance with additional ESXi hosts_' to 'The following table displays the improved performance of migrating 50 VMs with additional ESXi hosts.
- Maximum number of concurrent disk transfers section looks really great with the table and sub indexes.




[1] Maximum recommended number of VMs per host
- here we want to talk about Maximum recommended number of concurent disk migrations.  The way its currently written is that its focused on VMs to Host aspect and this is not correct as we can have hundreds of vms per host, how many we migrate concurrently is what your table is showing.  Let me know if unclear.

apinnick:
>     * '_Fast, dedicated networks for the migration_' should include line about datastore speeds, migrations are bound by esx datatstore speed, then network.

@mordechai  The performance document says this about datastore read speeds:

> Datastores read rates impact the total transfer times, it's important to ensure fast reads are possible from the esx datastore to the esx host.  Example in numbers: 800MB/s was the average read rate for vsphere endpoint, and 350-500MB/s on average for esx endpoint.  Speed of reads between endpoints are documented here:  (BZ 1957985)

There is a separate section about vSphere endpoint vs ESXi endpoint, including a table showing speed. I could add a line to the endpoint section about how the endpoints affect the datastore read speeds. I do not think it is necessary to repeat this section in the "fast network" recommendation when much of it is mentioned in the Endpoint section.

Re: max concurrent disk transfers/VMs -- yes, of course we mean concurrent migrations, not VMs existing on a host. It hadn't occurred to me that that point might be misunderstood. I will check each occurrence to make sure that we're only concerned about the migrating VMs/disks.

Everything else is clear.

mordechai:
Reviewed and looks good.
Thank you @apinnick

apinnick:
Thanks, @mordechai 
I have asked a couple writers to check it from a docs perspective. When they approve it, I will publish it on the Customer Portal and add a link to the Jira and close it.

apinnick:
Closing PR. Article published: https://access.redhat.com/articles/6242511

PR #224, opened 2021-10-20
BZ2012738: Changing precopy interval

apinnick:
https://bugzilla.redhat.com/show_bug.cgi?id=2012738

Previews:

https://deploy-preview-224--forklift-documentation.netlify.app/documentation/doc-migration_toolkit_for_virtualization/master/#changing-precopy-intervals-ui_forklift

### DIFF
--- a/documentation/modules/changing-precopy-intervals-ui.adoc
+++ b/documentation/modules/changing-precopy-intervals-ui.adoc
@@ -1,0 +1,18 @@
+// Module included in the following assemblies:
+//
+// * documentation/doc-Migration_Toolkit_for_Virtualization/master.adoc
+
+[id="changing-precopy-intervals-ui_{context}"]
+= Changing precopy intervals for warm migration
+
+The snapshots for the precopy stage of warm migration are created at one-hour intervals by default.
+
+You can change the snapshot interval by editing the `forklift-controller` deployment in the {ocp} web console.
+
+.Procedure
+
+. In the {ocp} web console, click *Workloads* > *Deployments*.
+. Click the `forklift-controller` project.
+. Click the *Environment* tab.
+. Click *Add more* under *Single values (env)*.
+. Enter `PRECOPY_INTERVAL` in the Name field, enter the value, and then click *Save*.
--- a/documentation/doc-Migration_Toolkit_for_Virtualization/master.adoc
+++ b/documentation/doc-Migration_Toolkit_for_Virtualization/master.adoc
@@ -28,6 +28,8 @@ You can migrate virtual machines from VMware vSphere or {rhv-full} to {virt} wit
 * link:https://access.redhat.com/articles/6242511[Performance recommendations for migrating from VMware vSphere to OpenShift Virtualization].
 * link:https://access.redhat.com/articles/6380311[Performance recommendations for migrating from Red Hat Virtualization to OpenShift Virtualization].
 
+include::modules/about-cold-warm-migration.adoc[leveloffset=+2]
+
 [id="prerequisites"]
 == Prerequisites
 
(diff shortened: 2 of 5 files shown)


apinnick:
@istein1  Please review. Thanks.

mguetta1:
@apinnick could you please add how to set the precopy_interval environment variable via UI? 
- Goto Workloads -> Deployments -> 'forklift-controller' -> Environment
- Add env variable named: PRECOPY_INTERVAL and value
- Save

apinnick:
> @apinnick could you please add how to set the precopy_interval environment variable via UI?
> 
>     * Goto Workloads -> Deployments -> 'forklift-controller' -> Environment
> 
>     * Add env variable named: PRECOPY_INTERVAL and value
> 
>     * Save

@mguetta1 Does the forklift-controller pod have to be restarted?

mguetta1:
@apinnick once pressing on 'save' the current forklift-controller pod is terminated and another one is created.
No need to manually restart

apinnick:
@mguetta1 Can you explain the use cases to me? 

Why would someone want to do it with an environmental variable rather than edit the forklift-controller deployment?

mguetta1:
@apinnick this is the same operation, to set the interval time between precopies, via UI.
In both cases, the user edits the forklift-controller deployment by adding an environment variable.
IN my opinion, since we have the ability to set the internal time using UI/CLI (related to MTV 2.2), we should provide both options

documentation/modules/changing-precopy-intervals-cli.adoc
```diff
@@ -0,0 +1,66 @@
...
+[id="changing-precopy-intervals-cli_{context}"]
+= Changing precopy intervals for warm migration
+
+The snapshots for the precopy stage of warm migration are created at one-hour intervals by default.
+
+You can change the snapshot interval by editing the `forklift-controller` deployment.
+
+.Prerequisites
+
+* You must enable link:https://kb.vmware.com/s/article/1020128[changed block tracking (CBT)] on the VMware virtual machines and virtual disks.
```
mguetta1, line 14: This is a general prerequisite in order to be able to warm migrate a VMware VM, not specific to change the interval time between precopies

documentation/modules/changing-precopy-intervals-ui.adoc
```diff
@@ -0,0 +1,22 @@
...
+[id="changing-precopy-intervals-ui_{context}"]
+= Changing precopy intervals for warm migration
+
+The snapshots for the precopy stage of warm migration are created at one-hour intervals by default.
+
+You can change the snapshot interval by editing the `forklift-controller` deployment in the {ocp} web console.
+
+.Prerequisites
+
+* You must enable link:https://kb.vmware.com/s/article/1020128[changed block tracking (CBT)] on the VMware virtual machines and virtual disks.
```
mguetta1, line 14: This is a general prerequisite in order to be able to warm migrate a VMware VM, not specific to change the interval time between precopies

apinnick:
> @apinnick this is the same operation, to set the interval time between precopies, via UI. In both cases, the user edits the forklift-controller deployment by adding an environment variable. IN my opinion, since we have the ability to set the internal time using UI/CLI (related to MTV 2.2), we should provide both options

@mguetta1 There needs to be a reason for documenting the same thing twice. 

@fdupont-redhat  Is there a reason why a user might prefer to do it by editing the forklift-controller deployment instead of adding a global variable?

documentation/modules/changing-precopy-intervals-cli.adoc
mguetta1: oh I see now that there is no dedicated section for warm migration.

apinnick: I have not yet decided where this module is going to go.

apinnick:
@mguetta1 I updated the module so that the UI method is documented. See the preview link in the PR description.

I'm still trying to decide whether putting it between "running migration plan" and "canceling migration plan" is the best place.

mguetta1 (approved):
Thanks @apinnick 
Looks good to me

PR #290, opened 2022-03-14
fix redundant callouts

stoobie:
Fixing redundant callouts like this:
image


stoobie:
@mguetta1 Please review the changes in `documentation/modules/canceling-migration-cli.adoc`

PR #295, opened 2022-03-21
Draft/WIP: Additional steps for upgrading to MTV 2.3

RichardHoch:
For MTV 2.3

Resolves https://bugzilla.redhat.com/show_bug.cgi?id=2063700

Describes additional steps that might need to be taken in an upgrade from MTV 2.2 to 2.3.

### DIFF
--- a/documentation/modules/upgrading-mtv-ui.adoc
+++ b/documentation/modules/upgrading-mtv-ui.adoc
@@ -37,7 +37,7 @@ $ {oc} get pod -n openshift-marketplace | grep <catalog_source> <1>
 <1> Specify the catalog source, for example, `redhat-operators`.
 ..  Delete the pod:
 +
-[source,terminal,subs=attributes+]
+[source,shell,subs=attributes+]
 ----
 $ {oc} delete pod -n openshift-marketplace <catalog_source_pod>
 ----
@@ -57,7 +57,7 @@ endif::[]
 
 . Verify that the `forklift-ui` pod is in a `Ready` state before you log in to the web console:
 +
-[source,terminal,subs="attributes+"]
+[source,shell,subs="attributes+"]
 ----
 $ {oc} get pods -n {namespace}
 ----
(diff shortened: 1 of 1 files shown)


documentation/modules/upgrading-mtv-ui.adoc
```diff
@@ -69,3 +69,9 @@ forklift-controller-788bdb4c69-mw268  2/2    Running  0         2m
 forklift-operator-6bf45b8d8-qps9v     1/1    Running  0         5m
 forklift-ui-7cdf96d8f6-xnw5n          1/1    Running  0         2m
 ----
++
```
istein1, line 72: Wanted to emphasize that there is no relation between possible added provider and possible added mapping.
These are 2 separate things that might have been added before upgrade and if users want to keep on using them,
they'll need to do some steps.

Also why there is a direction to MTV-2.2 doc /OCP-4.9?
This documentation change is for MTV-2.3.

documentation/modules/upgrading-mtv-ui.adoc
```diff
@@ -16,7 +16,7 @@ You must upgrade to the next release without skipping a release, for example, fr
 
 . In the {ocp-short} web console, click *Operators* -> *Installed Operators* -> *{operator-name-ui}* -> *Subscription*.
 
-. Change the update channel to *release-v2.2.0*.
+. Change the update channel to the correct release.
```
RichardHoch, line 19: @istein1 -- I made this change since the text referred to version 2.2. It it OK? Should I add "for example, *release-v.2.3.0" ?  I am guessing that the UI will say  *release-v.2.3.0"  ... Please check that everything is OK for version 2.3.

Do we need version by version upgrade instructions? I don't think so, but I want to raise the issue.

RichardHoch:
@istein1 I made the change you requested. Please review it also see my suggestion and question at line 19.  Thanks.

documentation/modules/upgrading-mtv-ui.adoc
RichardHoch: Change the text. The OCP version number is controlled by an attribute;  I changed its value.

istein1:
@RichardHoch , Looks good, thanks.

PR #319, opened 2022-04-05
Red callouts

RichardHoch:
Fixed redundant callout


documentation/modules/upgrading-mtv-ui.adoc
```diff
@@ -34,7 +34,9 @@ endif::[]
 ----
 $ {oc} get pod -n openshift-marketplace | grep <catalog_source> <1>
 ----
++
```
apinnick, line 37: Is the '+' intentional? It is not normally used for callouts.

PR #322, opened 2022-04-11
Replace [souce,shell..} with [source, terminal..] in repo

RichardHoch:
MTV 2.3

In some files, the block metadata began with [source,shell...] instead of the previous [source,terminal...]. After consulting with Avital Pinnick, I've decided to revert the changes in order to maintain consistency in the repo and the larger document set.


sbeskin-redhat:
@RichardHoch  LGTM

PR #352, opened 2023-02-22
Invent service json

RichardHoch:
MTV 4.2

Resolves https://issues.redhat.com/browse/MTV-470 by adding 2 steps to the procedure and making additional changes suggested in the PR.

Preview: http://file.emea.redhat.com/rhoch/invent_service_json/html-single/#retrieving-validation-service-json_mtv [steps 1-6]

### DIFF
--- a/documentation/modules/retrieving-validation-service-json.adoc
+++ b/documentation/modules/retrieving-validation-service-json.adoc
@@ -12,6 +12,13 @@ You can create a validation rule based on any attribute in the `"input"` key, fo
 
 .Procedure
 
+. Retrieve the routes for the project:
++
+[source,terminal]
+----
+oc get route -n openshift-mtv
+----
+
 . Retrieve the `Inventory` service route:
 +
 [source,terminal,subs="attributes+"]
@@ -19,11 +26,25 @@ You can create a validation rule based on any attribute in the `"input"` key, fo
 $ {oc} get route <inventory_service> -n {namespace}
 ----
 
+. Retrieve the access token:
++
+[source,terminal]
+----
+$ TOKEN=$(oc whoami -t)
+----
+
+. Trigger an HTTP GET request (for example, using Curl):
++
+[source,terminal]
+----
+$ curl -H "Authorization: Bearer $TOKEN" https://<inventory_service_route>/providers -k
+----
+
 . Retrieve the `UUID` of a provider:
 +
 [source,terminal]
 ----
-$ GET https://<inventory_service_route>/providers/<provider> <1>
+$ curl -H "Authorization: Bearer $TOKEN"  https://<inventory_service_route>/providers/<provider> -k <1>
 ----
 <1> Allowed values for the provider are `vsphere` and `ovirt`.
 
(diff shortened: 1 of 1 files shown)


ahadas (commented):
oh I just looked at the [preview](http://file.emea.redhat.com/rhoch/invent_service_json/html-single/#retrieving-validation-service-json_mtv) and I see that we have GET in the next steps, how about changing them as well?

RichardHoch:
@ahadas Do you mean replace GET in the shell command with _curl -H "Authorization: Bearer $TOKEN"_  or with just _curl -H_  ?

ahadas:
> @ahadas Do you mean replace GET in the shell command with _curl -H "Authorization: Bearer $TOKEN"_ or with just _curl -H_ ?

yes

RichardHoch:
@ahadas  I was not clear enough. Should I write
a) curl -H "Authorization: Bearer $TOKEN" 
or
b) url -H 

?

ahadas:
> @ahadas I was not clear enough. Should I write a) curl -H "Authorization: Bearer $TOKEN" or b) curl -H
> 
> ?

ah, the former (`curl -H "Authorization: Bearer $TOKEN"`)

documentation/modules/retrieving-validation-service-json.adoc
```diff
@@ -19,6 +19,20 @@ You can create a validation rule based on any attribute in the `"input"` key, fo
 $ {oc} get route <inventory_service> -n {namespace}
 ----
 
+. Retrieve the access token:
++
+[source,terminal]
+----
+$ TOKEN = $(oc whoami -t)
```
qiyuann, line 26: There shouldn't be space before and after the "="

documentation/modules/retrieving-validation-service-json.adoc
```diff
@@ -19,6 +19,20 @@ You can create a validation rule based on any attribute in the `"input"` key, fo
...
+----
+
+. Trigger an HTTP GET request (for example, using Curl):
++
+[source,terminal]
+----
+$ curl -H "Authorization: Bearer $TOKEN" https://<inventory_service_route>/providers
```
qiyuann, line 33: When I'm testing this command, I have to add -k, @ahadas, should it be added here?

RichardHoch: @qiyuann where do you add it?

qiyuann: just after curl, curl -k -H...

polinaag (commented):
Hi, I've tested the commands .  In my case it also required the '-k' flag to be added, like Qin already mentioned, so the command must be like that:
curl -H "Authorization: Bearer $TOKEN" https://<inventory_service_route>/providers -k
Another note: 
maybe it is worth to add before the step 1 , the initial step - oc get route -n openshift-mtv to get the routes for the project.

documentation/modules/retrieving-validation-service-json.adoc
RichardHoch: OK to all of the above.

polinaag (commented):
looks good to me

CarmiWisemon (commented):
LGTM

documentation/modules/retrieving-validation-service-json.adoc
ahadas: I don't remember having to use an insecure connection, but it could be that the reason for this was that I've triggered it from within the cluster, which is not a typical flow.. so that's ok to add the insecure flag

PR #355, opened 2023-03-14
MTV-488: Add disable hibernation to VM prerequisites

RichardHoch:
MTV 2.4

Resolves https://issues.redhat.com/browse/MTV-488 by including "disable hibernation" in the list of VMware migration prerequisites.

Preview: http://file.emea.redhat.com/rhoch/hibernate/html-single/#vmware-prerequisites_mtv (last item in list before the note)

### DIFF
--- a/documentation/modules/vmware-prerequisites.adoc
+++ b/documentation/modules/vmware-prerequisites.adoc
@@ -14,10 +14,16 @@ The following prerequisites apply to VMware migrations:
 * You must create a VMware Virtual Disk Development Kit (VDDK) image.
 * You must obtain the SHA-1 fingerprint of the vCenter host.
 * If you are migrating more than 10 VMs from an ESXi host in the same migration plan, you must increase the NFC service memory of the host.
+* It is strongly recommended to disable hibernation because {project-first} does not support migrating hibernated VMs.
+
+[IMPORTANT]
+====
+In the event of a power outage, data might be lost for a VM with disabled hibernation. However, if hibernation is not disabled, migration will fail
+====
 
 [NOTE]
 ====
-Neither {project-first} nor OpenShift Virtualization support conversion of Btrfs for migrating VMs from VMWare.
+Neither {project-short} nor OpenShift Virtualization support conversion of Btrfs for migrating VMs from VMWare.
 ====
 
 [discrete]


documentation/modules/vmware-prerequisites.adoc
```diff
@@ -14,6 +14,7 @@ The following prerequisites apply to VMware migrations:
 * You must create a VMware Virtual Disk Development Kit (VDDK) image.
 * You must obtain the SHA-1 fingerprint of the vCenter host.
 * If you are migrating more than 10 VMs from an ESXi host in the same migration plan, you must increase the NFC service memory of the host.
+* You must disable hibernation.
```
ahadas, line 17: @RichardHoch maybe we can be less strict here with something like:
Make sure there are no hibernated VMs. It is recommended to disable hiberation.

I just looked at t[he guidelines on how to disable hibernation in vsphere](https://docs.vmware.com/en/VMware-Horizon-7/7.13/published-desktops-applications/GUID-69E4CC65-7DD1-43B0-A522-8ABFEB9388EF.html) and users may want to avoid it when they read `When you make hibernation unavailable, hybrid sleep does not work. Users can lose data if a power loss occurs.`. It's not a strict prerequisite on our end but a recommendation so migrations won't fail on hibernated VMs

anarnold97: @RichardHoch & @ahadas thanks for the clarity.
I am concerned and want to ensure we cover this correctly.
I would like to include an IMPORT admonition, something like:

> [!IMPORTANT]
>
> It is recommended to disable hibernation.
> When you disable hibernation, hybrid sleep will not work. If there is a power loss, it is possible you could lose data.

Richard can make it look prettier, but would you be happy with the general context?

RichardHoch: I phrased it somewhat differently, with fewer details -- @ahadas I'm open to including more information. Let me know what you think is best for our users.

documentation/modules/vmware-prerequisites.adoc
```diff
@@ -14,10 +14,16 @@ The following prerequisites apply to VMware migrations:
 * You must create a VMware Virtual Disk Development Kit (VDDK) image.
 * You must obtain the SHA-1 fingerprint of the vCenter host.
 * If you are migrating more than 10 VMs from an ESXi host in the same migration plan, you must increase the NFC service memory of the host.
+* It is strongly recommended to disable hibernation because {project-first} does not support migrating hibernated VMs.
+
+[IMPORTANT]
+====
+If there is a power outage, data might be lost for a VM for which hibernation is disabled. However, if hibernation is not disabled, migration will fail.
```
anarnold97, line 21: @RichardHoch - i think this is very good; clear and concise with no ambiguity

@ahadas - you good with this?

ahadas: I'm still not sure about the last part - as far as I know we don't require to disable it.
@qiyuann could you please keep me honest - that migration succeeds for a VM for which hibernation is enabled but the VM is not in hibernated state?

.idea/workspace.xml
```diff
@@ -0,0 +1,40 @@
...
+    <task active="true" id="Default" summary="Default task">
+      <changelist id="03dfe95b-ac97-47c4-9aef-55705a77119f" name="Changes" comment="" />
+      <created>1678971888796</created>
+      <option name="number" value="Default" />
+      <option name="presentableId" value="Default" />
+      <updated>1678971888796</updated>
+    </task>
+    <servers />
+  </component>
+</project>
```
ahadas, line 40: this file was included  by mistake, right?

documentation/modules/vmware-prerequisites.adoc
ahadas: I spoke with Nisim yesterday and apparently I didn't understand it correctly - that "hibernation" setting is not about whether the VM can be hibernated but what's the operation that is executed when pressing the power-off button and as such it's not possible to shutdown a vm "normally" when hibernation is enabled. in that case, this phrasing makes sense. @nyoxi are you aware of this?

ahadas (approved):

documentation/modules/vmware-prerequisites.adoc
nyoxi: One cannot expect that hibernated VM will come up properly after being migrated. Hibernated OS does not tolerate too many hardware changes (Would you hibernate your OS before replacing mainboard in your bare metal machine?) and changes to the disk content are not necessarily safe either. Not to mention that you are likely to lose your saved RAM depending on the configuration of your OS. This also includes the fast boot/fast start feature of Windows where virt-v2v is not able to modify the filesystem and registry. What we should be telling people is to make sure their guest OS is properly shut down.

To extend on that, in IMS we had a document with "best practices" (Do we have something like that too?) where we were suggesting to not only shut down the guest but to perform a reboot before doing the shut down for migration. This is to make sure the OS is in a proper condition and to avoid any surprises after migration which could be mistaken as bugs in our tooling. There be partially installed apps waiting for restart or system updates in queue or just misconfigured services.

With that all being said, I don't thing we convey the necessary message to the users/admins.

nyoxi: > Not to mention that you are likely to lose your saved RAM depending on the configuration of your OS.

On a second thought, this should not really happen.

qiyuann (approved):

documentation/modules/vmware-prerequisites.adoc
CarmiWisemon, line 21: In the event of a power outage, data may be lost for a VM with disabled hibernation. However, if hibernation is not disabled, migration will fail.

CarmiWisemon (commented):
I made one suggestion.

CarmiWisemon (commented):
LGTM

PR #360, opened 2023-03-27
OpenStack prerequisites

RichardHoch:
MTV 4.2

Partially resolves https://issues.redhat.com/browse/MTV-426 by adding a new section, "OpenStack prerequisites," to the MTV guide.

Preview:  http://file.emea.redhat.com/rhoch/openstack_pre/html-single/#openstack-prerequisites_mtv

### DIFF
--- a/documentation/modules/common-attributes.adoc
+++ b/documentation/modules/common-attributes.adoc
@@ -10,6 +10,7 @@
 :operator: mtv-operator
 :operator-name-ui: Migration Toolkit for Virtualization Operator
 :operator-name: MTV Operator
+:osp: OpenStack
 :project-full: Migration Toolkit for Virtualization
 :project-short: MTV
 :project-first: {project-full} ({project-short})
--- a/documentation/modules/openstack-prerequisites.adoc
+++ b/documentation/modules/openstack-prerequisites.adoc
@@ -1,0 +1,28 @@
+// Module included in the following assemblies:
+//
+// * documentation/doc-Migration_Toolkit_for_Virtualization/master.adoc
+
+:_content-type: REFERENCE
+[id="openstack-prerequisites_{context}"]
+= OpenStack prerequisites
+
+The following prerequisites apply to {osp} migrations:
+
+* You must use a xref:compatibility-guidelines_{context}[compatible version] of {osp}.
+
+[IMPORTANT]
+====
+Migration using {osp} source providers is a Technology Preview feature only. Technology Preview features
+are not supported with Red Hat production service level agreements (SLAs) and
+might not be functionally complete. Red Hat does not recommend using them
+in production. These features provide early access to upcoming product
+features, enabling customers to test functionality and provide feedback during
+the development process.
+For more information about the support scope of Red Hat Technology Preview
+features, see https://access.redhat.com/support/offerings/techpreview/.
+====
+
+[NOTE]
+====
+Migration using {osp} source providers only supports VMs that use only Cinder volumes.
+====
--- a/documentation/doc-Migration_Toolkit_for_Virtualization/master.adoc
+++ b/documentation/doc-Migration_Toolkit_for_Virtualization/master.adoc
@@ -44,6 +44,7 @@ include::modules/storage-support.adoc[leveloffset=+2]
 include::modules/network-prerequisites.adoc[leveloffset=+2]
 include::modules/source-vm-prerequisites.adoc[leveloffset=+2]
 include::modules/rhv-prerequisites.adoc[leveloffset=+2]
+include::modules/openstack-prerequisites.adoc[leveloffset=+2]
 include::modules/vmware-prerequisites.adoc[leveloffset=+2]
 include::modules/creating-vddk-image.adoc[leveloffset=+3]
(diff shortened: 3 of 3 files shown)


.idea/workspace.xml
ahadas, line 1: this file needs to drop

documentation/modules/common-attributes.adoc
```diff
@@ -11,6 +11,7 @@
 :operator: mtv-operator
 :operator-name-ui: Migration Tookit for Virtualization Operator
 :operator-name: MTV Operator
+:osh: OpenStack
```
ahadas, line 14: for your consideration - we usually use the acronom `osp` (that stands for openstack platform), it would be better than `osh`

documentation/modules/openstack-prerequisites.adoc
```diff
@@ -0,0 +1,31 @@
...
+// * documentation/doc-Migration_Toolkit_for_Virtualization/master.adoc
+
+:_content-type: REFERENCE
+[id="openstack-prerequisites_{context}"]
+= OpenStack prerequisites
+
+The following prerequisites apply to {osh} migrations:
+
+* You must use a xref:compatibility-guidelines_{context}[compatible version] of {osh}.
+* You must have the {manager} CA certificate or the {manager} Apache CA certificate if the {manager} CA certificate was replaced by a third-party certificate on the Apache server.
```
ahadas, line 12: about the "You must have" - we'll support configuring an openstack provider without specifying a certificate (by selecting the option `Skip certificate validation (if checked, the provider's certificate won't be validated)`), so it's not a must-have

openstack doesn't have a central management unit, like rhv or vsphere, so we better avoid the use of "Manager" in this context. I'd also drop the `Apache` part - openstack is written in python and uses different tooling than RHV. in my opinion, it's ok to be a bit more vague here and say that the CA certificate for connecting to the source environment is needed, like we write for vSphere

documentation/modules/openstack-prerequisites.adoc
```diff
@@ -0,0 +1,31 @@
...
+in production. These features provide early access to upcoming product
+features, enabling customers to test functionality and provide feedback during
+the development process.
+For more information about the support scope of Red Hat Technology Preview
+features, see https://access.redhat.com/support/offerings/techpreview/.
+====
+
+[NOTE]
+====
+Migration using {osh} source providers only supports VMs whose volume snapshots are supported by {osh} Block Storage (Cinder). Migrations for VMs that use {osh} Swift at all or use only {osh} Glance for backend storage are not currently supported.
```
ahadas, line 30: @bennyz can you please check the correctness of this part?

documentation/modules/common-attributes.adoc
RichardHoch: good idea.

documentation/modules/openstack-prerequisites.adoc
bennyz: I think `glance for backend storage` is confusing, as glance is an image store, the idea here is convey that the VM should start from a volume rather than directly from glance, but the volume can be based on swift or glance

RichardHoch: @bennyz @ahadas  How about one of these? 

a) "Migration using {osp} source providers only supports VMs whose volume snapshots are supported by {osp} Block Storage (Cinder) or volumes based on {osp} Swift or {osp} Glance"  ?

b)  "Migration using {osp} source providers only supports VMs whose volumes  are supported by {osp} Block Storage (Cinder), {osp} Swift, or {osp} Glance."

bennyz: I think something like this would clear enough:
"Migration using {osp} source providers only supports VMs that uses only Cinder volumes"

RichardHoch: @bennyz -- Nice and clear -- thanks!

RichardHoch:
@bennyz @ahadas I think should do it...

documentation/modules/openstack-prerequisites.adoc
```diff
@@ -0,0 +1,28 @@
...
+//
+// * documentation/doc-Migration_Toolkit_for_Virtualization/master.adoc
+
+:_content-type: REFERENCE
+[id="openstack-prerequisites_{context}"]
+= OpenStack prerequisites
+
+The following prerequisites apply to {osp} migrations:
+
+* You must use a link:https://access.redhat.com/documentation/en-us/migration_toolkit_for_virtualization/2.3/html-single/installing_and_using_the_migration_toolkit_for_virtualization/index#compatibility-guidelines_mtv [compatible version of {osp}].
```
ahadas, line 11: need to remember to change this to 2.4

ahadas (approved):

documentation/modules/openstack-prerequisites.adoc
RichardHoch: fixed

RichardHoch:
@bennyz Please review this PR.

bennyz (approved):

qiyuann (approved):

PR #375, opened 2023-04-14
aa/MTV-475: adding 2.4.0 release notes

anarnold97:
@ahadas @qiyuann @mavital 

Adding a new clean version of the MTV 2.4.0 release notes without junk files and typos corrected. 
The other ones are such a mess, I thought it's best to start from scratch.
Please, could you do the necessary review? 
Thanks

PR #379, opened 2023-04-19
Cleanup: Adding providers

RichardHoch:
MTV 2.4

[no jira]: Updates Section 4.1, "Adding providers" to reflect changes to the UI or to cleanup aspects of the OCP plugin not covered in previous PRs.

Preview: section 4.1.1.1.1 in http://file.emea.redhat.com/rhoch/cleanup_chapter4/html-single/#adding-source-providers

### DIFF
--- a/documentation/modules/adding-virt-provider.adoc
+++ b/documentation/modules/adding-virt-provider.adoc
@@ -17,7 +17,7 @@ You can add {a-virt} destination provider to the {ocp} web console in addition t
 . In the {ocp} web console, click *Migration* -> *Providers for virtualization*.
 . Click *Create Provider*.
 . Select *{virt}* from the *Provider type* list.
-. Fill in the following fields:
+. Specify the following fields:
 
 * *Provider name*: Specify the provider name to display in the list of target providers.
 * *Kubernetes API server URL*: Specify the {ocp} cluster API endpoint.
--- a/documentation/modules/selecting-migration-network-for-vmware-source-provider.adoc
+++ b/documentation/modules/selecting-migration-network-for-vmware-source-provider.adoc
@@ -25,20 +25,14 @@ The source virtual disks are copied by a pod that is connected to the pod networ
 .Procedure
 
 . In the {ocp} web console, click *Migration* -> *Providers for virtualization*.
-. Click the *VMware* tab.
 . Click the host number in the *Hosts* column beside a provider to view a list of hosts.
 . Select one or more hosts and click *Select migration network*.
-. Select a *Network*.
-+
-You can clear the selection by selecting the default network.
+. Specify the following fields:
 
-. If your source provider is VMware, complete the following fields:
-* *ESXi host admin username*: Specify the ESXi host admin user, for example, `root`.
-* *ESXi host admin password*: Specify the ESXi host admin password.
+* *Network*: Network name
+* *ESXi host admin username*: For example, `root`
+* *ESXi host admin password*: Password
 
-. If your source provider is {rhv-full}, complete the following fields:
-* *Username*: Specify the {manager} user.
-* *Password*: Specify the {manager} password.
 . Click *Save*.
 . Verify that the status of each host is *Ready*.
 +
--- a/documentation/modules/adding-source-provider.adoc
+++ b/documentation/modules/adding-source-provider.adoc
@@ -28,7 +28,7 @@ endif::[]
 . Click *Create Provider*.
 ifdef::vmware[]
 . Select *VMware* from the *Provider type* list.
-. Fill in the following fields:
+. Specify the following fields:
 
 * *Provider name*: Name to display in the list of providers
 * *vCenter host name or IP address*: vCenter host name or IP address - if a certificate for FQDN is specified, the value of this field needs to match the FQDN in the certificate
@@ -41,7 +41,7 @@ ifdef::vmware[]
 endif::[]
 ifdef::rhv[]
 . Select *Red Hat Virtualization* from the *Provider type* list.
-. Fill in the following fields:
+. Specify the following fields:
 
 * *Provider name*: Name to display in the list of providers
 * *RHV Manager host name or IP address*: {manager} host name or IP address -  if a certificate for FQDN is specified, the value of this field needs to match the FQDN in the certificate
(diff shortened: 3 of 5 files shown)


documentation/modules/adding-virt-provider.adoc
```diff
@@ -7,7 +7,7 @@
 = Adding {a-virt} destination provider
 
 You can add {a-virt} destination provider to the {ocp} web console in addition to the default {virt} destination provider, which is the provider where you installed {project-short}.
-
+// Is this still accurate?
 .Prerequisites
```
RichardHoch, line 11: Please answer question in line 10...

RichardHoch:
@ahadas @qiyann Please review this PR.

documentation/modules/adding-virt-provider.adoc
ahadas: we're not supposed to require a service account token with cluster-admin privileges for the target provider but unless @qiyuann says they checked it (it's different than adding a provider with user that doesn't have cluster-admin privileges on the local OCP instance), I'd keep this part and the rest looks accurate

qiyuann: cluster-admin privileges is not a must. 
A service account token with privileges similar to rules defined in mtv-operator(forklift.konveyor.io, authentication.k8s.io, template.openshift.io are not required) could make remote migration work.

documentation/modules/selecting-migration-network-for-vmware-source-provider.adoc
```diff
@@ -28,17 +28,12 @@ The source virtual disks are copied by a pod that is connected to the pod networ
 . Click the *VMware* tab.
```
qiyuann, line 28:
```suggestion

```
There is no VMware tab

documentation/modules/adding-virt-provider.adoc
RichardHoch: @ahadas @qiyuann I wasn't clear enough. I meant is the description in line 9 still correct? I am going to leave the prerequisites as is unless you tell me otherwise.

qiyuann (commented):
Is the "Selecting a migration network for an OpenShift Virtualization provider" removed?
I didn't see there is such option on OpenShift Virtualization provider.

RichardHoch:
@ahadas @qiyann Please review this PR again.

RichardHoch:
@ahadas @qiyuann Can we set aside  "Selecting a migration network for an OpenShift Virtualization provider" for this PR?  That is, I suggest you review this PR as is and I will document "Selecting a migration network for an OpenShift Virtualization provide" in a different PR. That way, I can merge the changes in this PR before we publish on Sunday -- if what i have here is OK. What do you think?

qiyuann (approved):

documentation/modules/adding-virt-provider.adoc
```diff
@@ -7,7 +7,7 @@
 = Adding {a-virt} destination provider
 
 You can add {a-virt} destination provider to the {ocp} web console in addition to the default {virt} destination provider, which is the provider where you installed {project-short}.
-
+// Is this still accurate?
```
ahadas, line 10: don't forget to drop this

documentation/modules/adding-virt-provider.adoc
ahadas: yes, it is correct but this made us look at line 13 below - and according to @qiyuann 's feedback we can change it as we put it in #375 :  `You must have {a-virt} link:https://access.redhat.com/documentation/en-us/openshift_container_platform/{ocp-version}/html/authentication_and_authorization/using-service-accounts[service account token] with sufficient permissions on MTV resources (e.g., providers, plans) and Openshift Virtualization resources (e.g., VirtualMachine, DataVolume)`

RichardHoch:
@ahadas I made the changes you suggested. Please review again.

documentation/modules/adding-virt-provider.adoc
```diff
@@ -10,14 +10,16 @@ You can add {a-virt} destination provider to the {ocp} web console in addition t
 
 .Prerequisites
 
-* You must have {a-virt} link:https://access.redhat.com/documentation/en-us/openshift_container_platform/{ocp-version}/html/authentication_and_authorization/using-service-accounts[service account token] with `cluster-admin` privileges.
+* You must have {a-virt} link:https://access.redhat.com/documentation/en-us/openshift_container_platform/{ocp-version}/html/authentication_and_authorization/using-service-accounts[service account token] with sufficient permissions on {project-short} resources such as providers and plans and on Openshift Virtualization resources such as `VirtualMachine` and `DataVolume`.
```
qiyuann, line 13: For remote migration, there is no need to have permissions on MTV resources, generally, there is no MTV tool installed on the remote host.

```suggestion
* You must have {a-virt} link:https://access.redhat.com/documentation/en-us/openshift_container_platform/{ocp-version}/html/authentication_and_authorization/using-service-accounts[service account token] with sufficient permissions on Openshift Virtualization resources such as `VirtualMachine` and `DataVolume`.
```

ahadas: @qiyuann yes, that's true - migrations and plans were not good examples as they are created on the local OCP cluster, but we cannot limit this only to Openshift Virtualization resources because in case of imports from vSphere we also create the conversion pod there and we need to be able to create a secret - is there anything else actually?

ahadas: actually, maybe leaving it as it was ("service account token with `cluster-admin` privileges") is ok for now - we didn't change it in 2.4 but we didn't find the exact set of permissions required so we can change it when we have that set. @qiyuann WDYT?

qiyuann: It would requires permissions on `configmaps, events, namespaces, persistentvolumeclaims, persistvolumes, pods, serets, jobs, network-attachment-definitions, storageclasses, virtualmachines, datavolumes, securitycontextconstraints `

But I agree with leaving it as it was, which makes it much easier for users to apply.

RichardHoch: @ahadas @qiyuann So it should be "You must have {a-virt} link:https://access.redhat.com/documentation/en-us/openshift_container_platform/{ocp-version}/html/authentication_and_authorization/using-service-accounts[service account token] with `cluster-admin` privileges." Correct?

RichardHoch: @ahadas @qiyuann I reverted the changes to the prerequisite. Please take another look.

qiyuann (approved):

ahadas (approved):

PR #386, opened 2023-04-27
Cleanup upgrade

RichardHoch:
MTV 2.4

Partially resolves https://issues.redhat.com/browse/MTV-510 by updating "Chapter 7. Upgrading the Migration Toolkit for Virtualization" of the user guide.

Preview: http://file.emea.redhat.com/rhoch/cleanup_upgrade/html-single/#upgrading-mtv-ui_mtv

### DIFF
--- a/documentation/modules/upgrading-mtv-ui.adoc
+++ b/documentation/modules/upgrading-mtv-ui.adoc
@@ -7,11 +7,6 @@
 
 You can upgrade the {operator-name} by using the {ocp} web console to install the new version.
 
-[NOTE]
-====
-You must upgrade to the next release without skipping a release, for example, from 2.0 to 2.1 or from 2.1 to 2.2.
-====
-
 .Procedure
 
 . In the {ocp} web console, click *Operators* -> *Installed Operators* -> *{operator-name-ui}* -> *Subscription*.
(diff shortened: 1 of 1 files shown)


qiyuann (commented):
@ahadas, do we need to state that ForkliftController needs to be reinstalled after upgrading to 2.4 because of bug https://github.com/kubev2v/forklift/issues/314 ?

documentation/modules/upgrading-mtv-ui.adoc
```diff
@@ -64,12 +64,16 @@ $ {oc} get pods -n {namespace}
...
+forklift-volume-populator-controller-85d5cb64b6-mrlmc   1/1     Running   0          6m36s
 ----
 +
-. If you added VMware source providers to {project-short} 2.2, upgrading to {project-short} 2.3 changes the state of any VMware providers to Critical. To fix this problem, edit the VMwhere provider by adding a VDDK `init` image and verifying the certificate of the VMware provider. For more information see link:https://access.redhat.com/documentation/en-us/migration_toolkit_for_virtualization/2.2/html/installing_and_using_the_migration_toolkit_for_virtualization/migrating-vms-web-console#adding-source-provider_vmware[Addding a VMSphere source provider].
+. If you added VMware source providers to {project-short} 2.2, upgrading to {project-short} 2.3 changes the state of any VMware providers to Critical. To fix this problem, edit the VMware provider by adding a VDDK `init` image. For more information see link:https://access.redhat.com/documentation/en-us/migration_toolkit_for_virtualization/2.4/html/installing_and_using_the_migration_toolkit_for_virtualization/migrating-vms-web-console#adding-source-provider_vmware[Addding a VMSphere source provider].
```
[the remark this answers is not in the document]
RichardHoch: OK

ahadas, line 77: the `upgrading to {project-short} 2.3` part seems awkward since users that read the 2.4 documentation are not upgrading the 2.3. it's supposed to be the same when upgrading to 2.4 but instead of changing 2.3 to 2.4, I'd drop the version we upgrade to altogether - it should be obvious that we upgrade to the latest version
```suggestion
. If you upgrade from {project-short} 2.2 and have VMware source providers defined, the state of any VMware providers would change to Critical. To fix this problem, edit the VMware provider by adding a VDDK `init` image. For more information see link:https://access.redhat.com/documentation/en-us/migration_toolkit_for_virtualization/2.4/html/installing_and_using_the_migration_toolkit_for_virtualization/migrating-vms-web-console#adding-source-provider_vmware[Addding a VMSphere source provider].
```

documentation/modules/upgrading-mtv-ui.adoc
```diff
@@ -64,12 +64,16 @@ $ {oc} get pods -n {namespace}
...
 
-. If you mapped to NFS on the {ocp} destination provider in {project-short} 2.2, upgrading to {project-short} 2.3 invalidates the NFS mapping. To fix this problem, edit the `AccessModes` and `VolumeMode` parameters in the NFS storage profile. For more information, see link:https://access.redhat.com/documentation/en-us/openshift_container_platform/4.9/html-single/virtualization/index#virt-customizing-storage-profile_virt-creating-data-volumes[Customizing the storage profile].
+. If you mapped to NFS on the {ocp} destination provider in {project-short} 2.2, upgrading to {project-short} 2.3 invalidates the NFS mapping. To fix this problem, edit the `AccessModes` and `VolumeMode` parameters in the NFS storage profile. The same applies for upgrading to {project-short} 2.4. For more information, see link:https://access.redhat.com/documentation/en-us/openshift_container_platform/4.9/html-single/virtualization/index#virt-customizing-storage-profile_virt-creating-data-volumes[Customizing the storage profile].
```
ahadas, line 79: I'd go with a similar approach here as well so we won't need to include 2.5 later (at some point it would make sense to drop all things about 2.2)

ahadas:
> @ahadas, do we need to state that ForkliftController needs to be reinstalled after upgrading to 2.4 because of bug [kubev2v/forklift#314](https://github.com/kubev2v/forklift/issues/314) ?

nah, no need to - we'll release a fix quickly

RichardHoch:
@ahadas @qiyuann Please review this PR again. After talking with Arik an hour ago, I removed the note that said users must upgrade from y-stream to y-stream until they get to the current version (that's what that note meant, though it was phrased differently). It's no longer true.

documentation/modules/upgrading-mtv-ui.adoc
```diff
@@ -55,7 +50,7 @@ ifeval::["{build}" == "downstream"]
 See link:https://access.redhat.com/documentation/en-us/openshift_container_platform/{ocp-version}/html/operators/administrator-tasks#olm-approving-pending-upgrade_olm-upgrading-operators[Manually approving a pending upgrade] in the {ocp} documentation.
 endif::[]
 
-. Verify that the `forklift-ui` pod is in a `Ready` state before you log in to the web console:
+. Verify that the {project-short} pods are running:
```
ahadas, line 53: thinking about this again, previously it was necessary so one won't log-in the the OCP console when the route to the old UI doesn't exist - but now with the console-plugin it's no longer needed since the user gets a notification that the console has changed and needs to be refreshed, right? I think we can drop this part of verifying that the pods are running

ahadas: @yaacov ^

RichardHoch: OK, step, command and output removed.

RichardHoch:
@ahadas Made the change -- please review this PR again.

qiyuann (approved):

ahadas (approved):

PR #396, opened 2023-06-28
Change OpenStack to OpenShift

ahadas:
There's a typo in the "Adding destination providers" section: we refer to OpenStack instead of OpenShift when explaining that providing blank URL and Service account token in the add-provider dialog means the local cluster will be used. This PR replaces "local OpenStack cluster" with "local OpenShift cluster" to fix this.

### DIFF
--- a/documentation/modules/adding-virt-provider.adoc
+++ b/documentation/modules/adding-virt-provider.adoc
@@ -23,7 +23,7 @@ You can add {a-virt} destination provider to the {ocp} web console in addition t
 * *Kubernetes API server URL*: Specify the {ocp} cluster API endpoint.
 * *Service account token*: Specify the `cluster-admin` service account token.
 +
-If both *URL* and *Service account token* are left blank, the local {osp} cluster is used.
+If both *URL* and *Service account token* are left blank, the local {ocp-short} cluster is used.
 
 . Click *Create*.
 +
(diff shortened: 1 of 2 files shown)


RichardHoch:
@ahadas Looking at the attribute file common-attributes.adoc, I don't see :ocp-short: -- maybe you want to add it and define it as OpenShift. :ocp: is rendered as Red Hat OpenShift. Btw, looks like we are still showing the OCP version as 4.11 -- that's the :ocp-version: attribute. Should that be changed?

ahadas:
> @ahadas Looking at the attribute file common-attributes.adoc, I don't see :ocp-short: -- maybe you want to add it and define it as OpenShift. :ocp: is rendered as Red Hat OpenShift. Btw, looks like we are still showing the OCP version as 4.11 -- that's the :ocp-version: attribute. Should that be changed?

added `:ocp-short:` to common-attributes.adoc

about the :ocp-version:, we've released MTV 2.4 before OCP 4.13 was released so both 4.12 and 4.11 were supported. we'd need to change it when releasing 2.5

RichardHoch:
LGTM. @ahadas I understand from your comment that OCP = 4.11 is OK with you. If not, you can change the value.

RichardHoch (approved):

ahadas:
> LGTM. @ahadas I understand from your comment that OCP = 4.11 is OK with you. If not, you can change the value.

yeah, it's ok for 2.4.z
we'll need to change it for 2.5

RichardHoch:
@ahadas LGTM -- merge whenever you want.

PR #399, opened 2023-07-04
MTV-609: uninterpolated asciidoc variable, changing to OpenShift to fix

anarnold97:
Fixing small error in documentation - 

Line 22  change from:
*Warm Migration or migration to a remote {ocp-short} cluster:*
*Warm Migration or migration to a remote OpenShift cluster:*
image
I will fix completely in 2.5

### DIFF
--- a/documentation/modules/common-attributes.adoc
+++ b/documentation/modules/common-attributes.adoc
@@ -6,6 +6,7 @@
 :namespace: openshift-mtv
 :oc: oc
 :ocp: Red Hat OpenShift
+:ocp-name: OpenShift
 :ocp-short: OpenShift
 :ocp-version: 4.11
 :operator: mtv-operator
--- a/documentation/modules/virt-migration-workflow.adoc
+++ b/documentation/modules/virt-migration-workflow.adoc
@@ -19,7 +19,7 @@ You can use the detailed migration workflow to troubleshoot a failed migration.
 
 The workflow describes the following steps:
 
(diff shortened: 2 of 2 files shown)


anarnold97:
@ahadas & @qiyuann - just a quick fix. apologies but i will need to fix properly in 2.5. the attributes file link is missing from some module files and i need to work my way through and insert it in multiple files

qiyuann (approved):
LGTM

ahadas (commented):
@anarnold97 it was supposed to be replaced with OCP on downstream and OKD on upstream - I see that the term "OpenShift cluster" that you use here is pretty widespread (e.g., "Using Red Hat OpenShift Cluster Manager to work with your OpenShift clusters") so that's ok, but should we then later change ocp-short to be OpenShift on downstream?

and I see `{ocp-short}` in two more places in the documentation:
`Cold migration from RHV or OpenStack to the local {ocp-short} cluster:`
`Cold migration from VMWare to the local {ocp-short} cluster:`
shall we also fix them while on it?

anarnold97:
@ahadas - i have added a new attribute to see if this resolves the issue. I will publish to preview tomorrow and see if this resolves. If it does not, then I will do a bulk update to fix the issue.

anarnold97:
@ahadas & @qiyuann - are you ok if i merge and publish to stage to see if we have resolved? 
Thanks

anarnold97:
@ahadas @qiyuann 
Yeah Richard and me changing the variable worked. Thanks 

image

PR #402, opened 2023-07-18
New Providers UI

RichardHoch:
MTV 2.5

Resolves https://issues.redhat.com/browse/MTV-558 by adding new Providers to user guide and updating the procedures for creating providers to fit the new UI.

Previews: 
1. Source providers: https://file.emea.redhat.com/rhoch/new_providers_25/html-single/#adding-source-providers [4.1.1 through 4.1.1.5]
2. Destination providers: https://file.emea.redhat.com/rhoch/new_providers_25/html-single/#adding-destination-providers

Note: The requirement to have cluster-admin privileges for OpenShift Virtualization providers might change in a separate PR related to non-admin users.  Please ignore the issue here.

PR #411, opened 2023-08-15
OCP-OCP migration

RichardHoch:
MTV 2.5

Resolves https://issues.redhat.com/browse/MTV-575 by:
1. Adding an additional option to step 4 of "Creating a storage mapping."
2. Modifying the specification of the `vms` section of the `Plan` manifest for OCP source providers. 

Also adds a note about creating providers for future use to the CLI instructions. 

Previews:

1. https://file.emea.redhat.com/rhoch/ocp_ocp_mig/html-single/#creating-storage-mapping_mtv.
2. https://file.emea.redhat.com/rhoch/ocp_ocp_mig/html-single/#migrating-virtual-machines-from-cli [last note before prerequisites] 
3. https://file.emea.redhat.com/rhoch/ocp_ocp_mig/html-single/#migrating-virtual-machines-from-cli [step 7]

### DIFF
--- a/documentation/modules/creating-network-mapping.adoc
+++ b/documentation/modules/creating-network-mapping.adoc
@@ -17,15 +17,13 @@ You can create one or more network mappings by using the {ocp} web console to ma
 
 . In the {ocp} web console, click *Migration* -> *NetworkMaps for virtualization*.
 . Click *Create NetworkMap*.
-. Complete the following fields:
+. Specify the following fields:
 
 * *Name*: Enter a name to display in the network mappings list.
 * *Source provider*: Select a source provider.
 * *Target provider*: Select a target provider.
-+
-The *Source networks* and *Target namespaces/networks* text boxes become active.
 
-. Select a source network and a target namespace/network from the list.
+. Select a *Source network* and a *Target namespace/network*.
 . Optional: Click *Add* to create additional network mappings or to map multiple source networks to a single target network.
 . If you create an additional network mapping, select the network attachment definition as the target network.
 . Click *Create*.
--- a/documentation/modules/creating-storage-mapping.adoc
+++ b/documentation/modules/creating-storage-mapping.adoc
@@ -23,14 +23,14 @@ You can create a storage mapping by using the {ocp} web console to map source di
 * *Source provider*: Select a source provider.
 * *Target provider*: Select a target provider.
 
-. Map source disk storages to target storage classes as follows:
+. To create a storage mapping, click *Add* and map storage sources to target storage classes as follows:
 
 .. If your source provider is VMware vSphere, select a *Source datastore* and a *Target storage class*.
 .. If your source provider is {rhv-full}, select a *Source storage domain* and a *Target storage class*.
 .. If your source provider is {osp}, select a *Source volume type* and a *Target storage class*.
 .. If your source provider is a set of one or more OVA files, select a *Source* and a *Target storage class* for the dummy storage that applies to all virtual disks within the OVA files.
-
-. Optional: Click *Add* to create additional storage mappings or to map multiple source disk storages to a single target storage class.
+.. If your storage provider is {virt}. select a *Source storage class* and a *Target storage class*.
+.. Optional: Click *Add* to create additional storage mappings, including mapping multiple storage sources to a single target storage class.
 . Click *Create*.
 +
 The mapping is displayed on the *StorageMaps* page.
--- a/documentation/modules/migrating-virtual-machines-cli.adoc
+++ b/documentation/modules/migrating-virtual-machines-cli.adoc
@@ -253,34 +253,39 @@ spec:
     destination:
       name: <destination_cluster>
       namespace: {namespace}
-  map:
-    network: <3>
-      name: <network_map> <4>
+  map: <3>
+    network: <4>
+      name: <network_map> <5>
       namespace: {namespace}
-    storage:
-      name: <storage_map> <5>
+    storage: <6>
+      name: <storage_map> <7>
       namespace: {namespace}
   targetNamespace: {namespace}
-  vms: <6>
-    - id: <source_vm> <7>
+  vms: <8>
+    - id: <source_vm> <9>
     - name: <source_vm>
-      hooks: <8>
+      namespace: {namespace} <10>
+      hooks: <11>
         - hook:
             namespace: {namespace}
-            name: <hook> <9>
-          step: <step> <10>
+            name: <hook> <12>
+          step: <step> <13>
 EOF
 ----
 <1> Specify the name of the `Plan` CR.
 <2> Specify whether the migration is warm or cold. If you specify a warm migration without specifying a value for the `cutover` parameter in the `Migration` manifest, only the precopy stage will run.
-<3> You can add multiple network mappings.
-<4> Specify the name of the `NetworkMap` CR.
-<5> Specify the name of the `StorageMap` CR.
-<6> You can use either the `id` _or_ the `name` parameter to specify the source VMs.
-<7> Specify the VMware VM MOR, {rhv-short} VM UUID or the {osp} VM UUID.
-<8> Optional: You can specify up to two hooks for a VM. Each hook must run during a separate migration step.
-<9> Specify the name of the `Hook` CR.
-<10> Allowed values are `PreHook`, before the migration plan starts, or `PostHook`, after the migration is complete.
+<3> Specify only one network map and one storage map per plan.
+<4> Specify a network mapping even if the VMs to be migrated are not assigned to a network. The mapping can be empty in this case.
+<5> Specify the name of the `NetworkMap` CR.
+<6> Specify a storage mapping even if the VMs to be migrated are not assigned with disk images. The mapping can be empty in this case.
+<7> Specify the name of the `StorageMap` CR.
+<8> For all source providers except for {virt}, you can use either the `id` _or_ the `name` parameter to specify the source VMs. +
+{virt} source provider only: You can use only the `name` parameter, not the `id.` parameter to specify the source VMs.
+<9> Specify the VMware VM MOR, {rhv-short} VM UUID or the {osp} VM UUID.
+<10> {virt} source provider only.
+<11> Optional: You can specify up to two hooks for a VM. Each hook must run during a separate migration step.
+<12> Specify the name of the `Hook` CR.
+<13> Allowed values are `PreHook`, before the migration plan starts, or `PostHook`, after the migration is complete.
 
 . Create a `Migration` manifest to run the `Plan` CR:
 +


documentation/modules/creating-storage-mapping.adoc
```diff
@@ -28,6 +28,7 @@ You can create a storage mapping by using the {ocp} web console to map source di
 .. If your source provider is VMware, select a *Source datastore* and a *Target storage class*.
 .. If your source provider is {rhv-full}, select a *Source storage domain* and a *Target storage class*.
 .. If your source provider is {osp}, select a *Source volume type* and a *Target storage class*.
+.. If your storage provider is {virt}. select a *Source provider* and a *Target provider*.
```
ahadas, line 31: should be `Source storage class` and `Target storage class`, does it really refer to "provider" at the moment?

documentation/modules/migrating-virtual-machines-cli.adoc
```diff
@@ -258,23 +258,26 @@ spec:
...
+          step: <step> <11>
 EOF
 ----
 <1> Specify the name of the `Plan` CR.
 <2> Specify whether the migration is warm or cold. If you specify a warm migration without specifying a value for the `cutover` parameter in the `Migration` manifest, only the precopy stage will run.
 <3> You can add multiple network mappings.
 <4> Specify the name of the `NetworkMap` CR.
 <5> Specify the name of the `StorageMap` CR.
-<6> You can use either the `id` _or_ the `name` parameter to specify the source VMs.
+<6> For all source providers except for {virt}, you can use either the `id` _or_ the `name` parameter to specify the source VMs. +
```
bennyz, line 274: `namespace` needs to be specified as well

documentation/modules/creating-storage-mapping.adoc
RichardHoch: ocp_ocp_storage map
This is what I saw using OCP providers @qiyuann created for me -- but maybe i did something wrong? Xin, what are you seeing for this step?

sgratch: @RichardHoch two comments:

-  I suggest adding another step between step 3 to 4 - 

   After following step 3, the dialog looks as follows:

   Screenshot from 2023-08-16 21-23-17

   There are 2 options for the user now:
    a. If he wants to create an **empty** mapping (i.e. declare the source and target providers but leave the specific storage mappings empty for future use) then jump to step 6 by clicking the **Create** button.

     b. If he wants to set the mapping(s) instead of leaving it empty, then in step 4 we need to add that the user should click the **Add** button to create the  first  storage mappings. 

   After clicking the **Add** button in (b), the dialog looks like this:

      Screenshot from 2023-08-16 21-31-50

    **Please note** that this comment is not relevant specifically to OCP - OCP migration but to all providers and also seems relevant to the network mapping dialog (`step 4.3 Creating a network mapping`).

- Regarding step 4.d, it should be `Source storage class` and `Target storage class` as mentioned by Arik. 
  Currently there is a [bug](https://github.com/kubev2v/forklift-console-plugin/issues/691) since the `Source` label appears for the source instead of `Source storage class`. We'll fix it.

qiyuann: > This is what I saw using OCP providers @qiyuann created for me -- but maybe i did something wrong? Xin, what are you seeing for this step?

The Source/Target storage class could appear when the `Add` button is clicked.

documentation/modules/migrating-virtual-machines-cli.adoc
ahadas: yeah, that's (8)
@RichardHoch I think we need to consider replacing this part with an example per-provider as it starts to become complicated to have many exceptions per-item

RichardHoch: @ahadas @anarnold97 I think having an example per provider is a good idea. Can it wait until after GA?

RichardHoch:
@sgratch @ahadas Please review this PR again. Sharon thanks for the info about the empty mapping.
@bennyz I think Arik answered your question, but if you think callouts 6 and 8 of step 7, the `Plan` manifest are not clear enough, please let me know.

documentation/modules/migrating-virtual-machines-cli.adoc
anarnold97: @RichardHoch - I will have to defer to Arik, but I think the example is a good idea.
Do you think you could get it done in your post-release clean-up to minimize the delay?
Thanks

RichardHoch: @ahadas @anarnold97 We are talking about this section https://access.redhat.com/documentation/en-us/migration_toolkit_for_virtualization/2.4/html-single/installing_and_using_the_migration_toolkit_for_virtualization/index#migrating-virtual-machines-from-cli

@ahadas  Did you mean we should have examples per step (that is, all the examples of Secret manifest in step 1, all the the examples of Provider manifests in step 2, etc.) or instructions per provider type (that is RHV manifests from start to finish in one section, vSphere from start to finish in another)? 

In theory, i should be able to create all the manifests from what's written here and have QE check them. 

In reply to another PR, Arik suggested that we reorganize the document by provider, as in...

1. vSphere migrations
  a. UI
  b. CLI
  
2. RHV migrations
 a. UI
 b. CLI 
 
That might be a good long-term idea.

ahadas: > @ahadas Did you mean we should have examples per step (that is, all the examples of Secret manifest in step 1, all the the examples of Provider manifests in step 2, etc.) or instructions per provider type (that is RHV manifests from start to finish in one section, vSphere from start to finish in another)?

instructions per provider type

> 
> In theory, i should be able to create all the manifests from what's written here and have QE check them.
> 
> In reply to another PR, Arik suggested that we reorganize the document by provider, as in...
> 
> 1. vSphere migrations
>    a. UI
>    b. CLI
> 2. RHV migrations
>    a. UI
>    b. CLI

yeah, in those (b) sections

documentation/modules/creating-network-mapping.adoc
```diff
@@ -17,13 +22,11 @@ You can create one or more network mappings by using the {ocp} web console to ma
 
 . In the {ocp} web console, click *Migration* -> *NetworkMaps for virtualization*.
 . Click *Create NetworkMap*.
-. Complete the following fields:
+. Specify the following fields:
 
-* *Name*: Enter a name to display in the network mappings list.
+* *Name*: Enter a name to display in the network mappings list. For example, `Default network map` or `Empty network mapping`.
```
ahadas, line 27: I'd skip the "Empty network mapping" part, see my previous comment

documentation/modules/creating-storage-mapping.adoc
```diff
@@ -8,6 +8,11 @@
 
 You can create a storage mapping by using the {ocp} web console to map source disk storages to {virt} storage classes.
 
+[NOTE]
+====
+You can specify the source and target providers but leave the specific storage mappings empty for future use if VMs currently have no storages. This kind of mapping is called an _empty mapping_.
```
ahadas, line 13: same here - there's nothing special about these empty mappings, they are just a concrete case
in https://github.com/kubev2v/forklift-console-plugin/pull/498 we enabled to specify empty mappings since VMs in RHV that only have LUN disks, have no storage domains to map to storage classes - I'd put a note in the UI and CLI sections that talk about plan creation, that say that storage/network mappings should always be specified when creating a plan even when the migrated VMs have no network/storage

documentation/modules/creating-storage-mapping.adoc
```diff
@@ -19,17 +24,22 @@ You can create a storage mapping by using the {ocp} web console to map source di
 . Click *Create StorageMap*.
 . Specify the following fields:
 
-* *Name*: Enter a name to display in the storage mappings list.
+* *Name*: Enter a name to display in the storage mappings list. For example, `Default storage map` or `Empty storage mapping`.
```
ahadas, line 27: also here I'd drop the "Empty storage mapping"

documentation/modules/creating-storage-mapping.adoc
```diff
@@ -19,17 +24,22 @@ You can create a storage mapping by using the {ocp} web console to map source di
...
 * *Source provider*: Select a source provider.
 * *Target provider*: Select a target provider.
 
-. Map source disk storages to target storage classes as follows:
+. To create a storage mapping, do one of the following:
+
+** To create an empty mapping, click *Create*.
```
ahadas, line 33: and drop this part

documentation/modules/creating-storage-mapping.adoc
ahadas: actually that note should probably be only in the CLI section since users cannot skip creating/selecting mappings in the UI

documentation/modules/creating-network-mapping.adoc
```diff
@@ -8,6 +8,11 @@
 
 You can create one or more network mappings by using the {ocp} web console to map source networks to {virt} networks.
 
+[NOTE]
+====
+You can specify the source and target providers for future use if VMs currently have no storages. This kind of mapping is called an _empty mapping_.
```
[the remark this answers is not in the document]
RichardHoch: Removed.

documentation/modules/creating-network-mapping.adoc
RichardHoch: Removed sentence. it was only there to mention "empty network mapping".

documentation/modules/creating-storage-mapping.adoc
RichardHoch: Removed sentence as above.

qiyuann (approved):
LGTM

documentation/modules/creating-storage-mapping.adoc
sgratch: BTW, choosing this solution was easier for us than totally skipping on the mapping steps in the wizard, but we can consider doing that with the refactored new dialogs if required.

sgratch:
@RichardHoch @ahadas please see comment https://github.com/kubev2v/forklift-documentation/pull/411#discussion_r1318699643

documentation/modules/creating-storage-mapping.adoc
```diff
@@ -23,14 +23,14 @@ You can create a storage mapping by using the {ocp} web console to map source di
 * *Source provider*: Select a source provider.
 * *Target provider*: Select a target provider.
 
-. Map source disk storages to target storage classes as follows:
+. To create a storage mapping, click *Add* and map source disk storages to target storage classes as follows:
```
anarnold97, line 26:
```suggestion
. To create a storage mapping, click *Add* and map source disk storage to target storage classes as follows:
```

Unsure about this one… I think storage might convey singular or plural

documentation/modules/creating-storage-mapping.adoc
```diff
@@ -23,14 +23,14 @@ You can create a storage mapping by using the {ocp} web console to map source di
...
 
 .. If your source provider is VMware vSphere, select a *Source datastore* and a *Target storage class*.
 .. If your source provider is {rhv-full}, select a *Source storage domain* and a *Target storage class*.
 .. If your source provider is {osp}, select a *Source volume type* and a *Target storage class*.
 .. If your source provider is a set of one or more OVA files, select a *Source* and a *Target storage class* for the dummy storage that applies to all virtual disks within the OVA files.
-
-. Optional: Click *Add* to create additional storage mappings or to map multiple source disk storages to a single target storage class.
+.. If your storage provider is {virt}. select a *Source storage class* and a *Target storage class*.
+.. Optional: Click *Add* to create additional storage mappings or to map multiple source disk storages to a single target storage class.
```
anarnold97, line 33:
```suggestion
.. Optional: Click *Add* to create additional storage mappings or to map multiple source disk storage to a single target storage class.
```
*???* - unsure. WDYT
see above

documentation/modules/migrating-virtual-machines-cli.adoc
```diff
@@ -21,6 +21,11 @@ include::snippet_ova_tech_preview.adoc[]
 Migration using {osp} source providers only supports VMs that use only Cinder volumes.
 ====
 
+[NOTE]
+====
+You can create `Provider` manifests for future use if the virtual machines that they include currently have no storages.
```
anarnold97, line 26:
```suggestion
Create manifests for future use if the virtual machines they include currently lack storage.
```

this is a tenuous comment - please ignore if you feel so minded

documentation/modules/migrating-virtual-machines-cli.adoc
```diff
@@ -264,23 +269,26 @@ spec:
...
+          step: <step> <11>
 EOF
 ----
 <1> Specify the name of the `Plan` CR.
 <2> Specify whether the migration is warm or cold. If you specify a warm migration without specifying a value for the `cutover` parameter in the `Migration` manifest, only the precopy stage will run.
 <3> You can add multiple network mappings.
 <4> Specify the name of the `NetworkMap` CR.
 <5> Specify the name of the `StorageMap` CR.
-<6> You can use either the `id` _or_ the `name` parameter to specify the source VMs.
+<6> For all source providers except for {virt}, you can use either the `id` _or_ the `name` parameter to specify the source VMs. +
```
anarnold97, line 285: Nice... the only problem is that the others are not as well written!

documentation/modules/creating-storage-mapping.adoc
ahadas: yeah, github also doesn't like `storages`..
how about:
```suggestion
. To create a storage mapping, click *Add* and map disk storage sources to target storage classes as follows:
```

documentation/modules/creating-storage-mapping.adoc
ahadas: > ... but we enhanced that as a feature to say that you can always create a plan with unmapped networks/storage, just to avoid blocking the plan creation, and it can be filled out later on once relevant.

to me this approach seems to go against the recent changes we've made in the UI to improve the user experience - as a user, I'd prefer the UI to block me from posting invalid input rather than posting it and then set a failure on the entity that would block me from going forward. in this case, I'd prefer to cancel the plan creation, create a mapping and then create the plan. btw, does it even work when the mapping is updated? AFAIR the UI clones the template mapping - do you go over the clones as well? @sgratch

RichardHoch: @sgratch @ahadas It sounds like there are two issues here. First, the phrase "empty mapping" isn't useful, so I can just remove the second sentence in the note. Second, do we want to call attention to this functionality? If not, it makes sense to remove the note altogether. But...maybe we want to leave the first sentence of the note in but add a warning such as "Be sure the mapping is still valid before attempting to implement it"? [Phrasing suggestions are welcome.]

ahadas: @RichardHoch +1 on dropping the "empty mapping" part
about the other part - in general, our documentation should reflect how simple it is to migrate using MTV rather than describing all the features (and in this case, I'm not even sure if this feature makes sense as it raises few questions, e.g., what if I create a plan with "valid" mappings and then someone changes it and make the plan invalid - this is something we should avoid)

RichardHoch:
@ahadas @sgratch Please review this PR again.

documentation/modules/migrating-virtual-machines-cli.adoc
```diff
@@ -21,6 +21,11 @@ include::snippet_ova_tech_preview.adoc[]
 Migration using {osp} source providers only supports VMs that use only Cinder volumes.
 ====
 
+[NOTE]
+====
+You can create `Provider` manifests for future use if the virtual machines that they include currently lack storage.
```
ahadas, line 26: I still don't understand what this means.. that's related to the empty mappings we talked about before, right?
in that case, I'd suggest to add to `network: <3>` below a note that says something like: network mapping should be specified also if the vms to be migrated don't have networks, the mapping can be empty in this case.

btw, I'm looking at `network: <3>` below and seems we need to change the current description of this item ("You can add multiple network mappings") - the plan can point to a single network (or storage) mapping

ahadas (commented):
besides what I commented on, the changes look good

documentation/modules/migrating-virtual-machines-cli.adoc
RichardHoch: @ahadas  about changing "You can add multiple network mappings" Are you saying we allow one network map per plan? If so, do I need to state this for storage below (no callout now).

ahadas: @RichardHoch correct. the network map can map multiple source networks to destination networks

RichardHoch:
@ahadas Made the changes you suggested. Please review.

documentation/modules/migrating-virtual-machines-cli.adoc
```diff
@@ -253,34 +253,38 @@ spec:
...
-<3> You can add multiple network mappings.
-<4> Specify the name of the `NetworkMap` CR.
-<5> Specify the name of the `StorageMap` CR.
-<6> You can use either the `id` _or_ the `name` parameter to specify the source VMs.
-<7> Specify the VMware VM MOR, {rhv-short} VM UUID or the {osp} VM UUID.
-<8> Optional: You can specify up to two hooks for a VM. Each hook must run during a separate migration step.
-<9> Specify the name of the `Hook` CR.
-<10> Allowed values are `PreHook`, before the migration plan starts, or `PostHook`, after the migration is complete.
+<3> Specify only one network map and one storage map per plan.
+<4> Specify a network mapping even if the VMs to be migrated are not assigned to a network. The mapping can be empty in this case.
```
ahadas, line 278: cool, and we need a similar note for the storage mapping

RichardHoch: Done.

RichardHoch:
@ahadas Made teh change you suggested. Please review.

documentation/modules/creating-storage-mapping.adoc
sgratch: @ahadas @RichardHoch 

Regarding the current functionality discussion:
> to me this approach seems to go against the recent changes we've made in the UI to improve the user experience - as a user, I'd prefer the UI to block me from posting invalid input rather than posting it and then set a failure on the entity that would block me from going forward. in this case, I'd prefer to cancel the plan creation, create a mapping and then create the plan. 

Correct,  please remember that this plan wizard is a legacy code. For 2.5.0, the current wizard behavior avoids blocking plan creation for invalid/wrong/empty mappings, since the assumption was that we didn't want to block it. So the plan is created with an error status and a detailed tooltip of, e.g:

Screenshot from 2023-09-11 16-17-51

We'll need to handle that as part of the refactored new dialog/wizard. I'll suggest opening an issue on this.


> btw, does it even work when the mapping is updated? AFAIR the UI clones the template mapping - do you go over the cloned mappings as well? @sgratch

No we don't. It will be left as is.
In case you edit the plan, you might get a warning on the mapping step that the previous mapping is invalid, but it won't be related to the original template and it's not auto updated once the template is updated.

RichardHoch:
@qiyuann Please review this PR.

sgratch (commented):
LGTM

~~- One small issue to consider: 
Do we want to consider adding a note regarding the special case of  "Host cluster" to  "Host cluster" mapping/plan?  
  By "Host cluster", I refer to the default ocp provider. 
  We do support it, not sure it's recommended except for testing....~~

  Update: I see that this was already added as part of https://github.com/kubev2v/forklift-documentation/pull/402/files#diff-34cd0c83438a7adb7c57b031ab1c68de180e668e14dc882e26f4b37cee901f76R61 

- Please see my last comment regarding the current ui functionality: https://github.com/kubev2v/forklift-documentation/pull/411#discussion_r1321550175

sgratch (approved):

qiyuann (approved):

PR #419, opened 2023-09-05
Provider names

RichardHoch:
MTV 2.5.0

 Resolves https://issues.redhat.com/browse/MTV-684 by adding a note about provider names to the UI instructions.

Previews:
1. https://file.emea.redhat.com/rhoch/prov_names/html-single/#adding-providers [Note added per provider]

### DIFF
--- a/documentation/modules/adding-source-provider.adoc
+++ b/documentation/modules/adding-source-provider.adoc
@@ -98,6 +98,8 @@ ifdef::rhv[]
 . Specify the following fields:
 
 * *Provider resource name*: Name of the source provider
++
+include::snip-provider-names.adoc[]
 * *URL*: API endpoint of the {rhv-short} Manager on which the source VM is mounted
 * *Username*: Username
 * *Password*: Password
@@ -115,13 +117,17 @@ ifdef::ova[]
 . Specify the following fields:
 
 * *Provider resource name*: Name of the source provider
++
+include::snip-provider-names.adoc[]
 * *URL*: URL of the NFS file share that serves the OVA
 endif::[]
 ifdef::ostack[]
 . Click *OpenStack*.
 . Specify the following fields:
 
-* *Provider resource name*: Name of the source provider.
+* *Provider resource name*: Name of the source provider
++
+include::snip-provider-names.adoc[]
 * *URL*: {osp} Identity (Keystone) endpoint, for example, `http://controller:5000/v3`.
 * *Authentication type*: Choose one of the following methods of authentication and supply the information related to your choice. For example, if you choose *Application credential ID* as the authentication type, the *Application credential ID* and the *Application credential secret* fields become active, and you need to supply the ID and the secret.
 
@@ -165,6 +171,8 @@ ifdef::cnv[]
 . Specify the following fields:
 
 * *Provider resource name*: Name of the source provider
++
+include::snip-provider-names.adoc[]
 * *URL*: Endpoint of the API server
 * *Service account bearer token*: Token for a service account with `cluster-admin` privileges
 +
@@ -175,6 +183,8 @@ ifdef::cnv2[]
 . Specify the following fields:
 
 * *Provider resource name*: Name of the source provider
++
+include::snip-provider-names.adoc[]
 * *URL*: Endpoint of the API server
 * *Service account bearer token*: Token for a service account with `cluster-admin` privileges
 +
--- a/documentation/modules/snip-provider-names.adoc
+++ b/documentation/modules/snip-provider-names.adoc
@@ -1,0 +1,7 @@
+:_content-type: SNIPPET
+
+[NOTE]
+====
+Provider names must start with an alphabetic character and cannot include the following character: +
+`.`
+====


RichardHoch:
@ahadas @yaacov @sgratch Please review -- if another SME needs to check the CLI section, please tag them.
@qiyuann Please review the PR as well.

sgratch (commented):
This LGTM.

Worth adding the following doc comment to the `plan name` , `network mapping name` and `storage mapping name` as follows:
The xxx name  cannot include the following character: `.`
(starting with an alphanumeric char is supported for those entities)

ahadas:
> hmm, sorry to raise this issue so late in the proccess: so, all the time (since we have UI) the UI did not let users enter numbers as first letter in names and no one had an issue with that until now?
> 
> @ahadas @qiyuann why do we need to document it at all (and only for the time span between 2.5.0 and 2.5.1) ?

did we have this restriction also with the standalone UI?

qiyuann:
> hmm, sorry to raise this issue so late in the proccess: so, all the time (since we have UI) the UI did not let users enter numbers as first letter in names and no one had an issue with that until now?
> 
@yaacov, do you mean only the provider name doesn't support a name starts with a number? I saw the plan name, storage map name, and network map name could start with a number.

qiyuann (commented):
On UI, when an invalid provider name is entered, it only shows a general message "Error: Name is required and must be a unique and valid Kubernetes name." , so better to add a note in the UI part of creating provider that the provider name should start with an alphabet character and not include '.'

yaacov:
> @yaacov did we have this restriction also with the standalone UI?
> I agree that it's insignificant but if we limit something that is possible to achieve using the CLI then it's just for the sake of avoid getting issues in the form of "why can't I create via he UI a plan with name X that I can create using the CLI"

that is what I understand from https://github.com/kubev2v/forklift-documentation/pull/419#pullrequestreview-1611203814

but looking more carefully at the stand alone code [1], they allow the name start with a number but didn't allow "." dots,
the plot thickens :-)

[1] https://github.com/kubev2v/forklift-console-plugin/blob/bacd7d98885474858929177acc3b02fc83d3c634/packages/legacy/src/common/constants.ts#L110

documentation/modules/creating-migration-plan.adoc
```diff
@@ -24,6 +24,12 @@ You can configure a hook to run an Ansible playbook or custom container image du
 . Specify the following fields:
 
 * *Plan name*: Enter a migration plan name to display in the migration plan list.
++
+[NOTE]
+====
+Plan names cannot include the following character: +
```
qiyuann, line 30: I think it's not necessary to add a note here, this part is creating plans on UI, there is a clear error message telling users what should be contained in the plan name, see "Plan name can only contain lowercase alphanumeric characters and dashes (-), and must start and end with an alphanumeric character"

documentation/modules/migrating-virtual-machines-cli.adoc
```diff
@@ -80,22 +80,24 @@ $ cat << EOF | {oc} apply -f -
...
-    namespace: {namespace}
+    name: <secret> <5>
+    namespace: <namespace>
 EOF
 ----
-<1> Allowed values are `ovirt`, `vsphere`, and `openstack`.
-<2> Specify the API end point URL, for example, `https://<vCenter_host>/sdk` for vSphere, `https://<engine_host>/ovirt-engine/api/` for {rhv-short}, or `https://<identity_service>/v3` for {osp}.
-<3> VMware only: Specify the VDDK image that you created.
-<4> Specify the name of provider `Secret` CR.
+<1> Provider names cannot include the following character: +
```
qiyuann, line 95: On UI, the Provider name can't include '.', and can't start with a number. Better to add 'can't start with a number' here as well.

yaacov:
> @yaacov, do you mean only the provider name doesn't support a name starts with a number? I saw the plan name, storage map name, and network map name could start with a number.

@qiyuann , correct, plans and mappings have different restrictions, in https://github.com/kubev2v/forklift-documentation/pull/419#pullrequestreview-1611203814 it was suggested to add the comment to plans and mappings, this imply that the restrictions apply to plans and mappings too, and is incurrect as the restrictions are different, they allow numbers but do not allow dots.

yaacov:
IMHO the comment about the name creates more confusion than helps us prevent "why can't I create via he UI a plan with name X that I can create using the CLI" questions.

we can add a comment about a known bug  ( https://github.com/kubev2v/forklift-console-plugin/issues/712 ) instead of trying to hide this bug as a "feature" by adding it to the docs as such

ahadas:
> EDIT @ahadas i copied your remark, it is providers not plans :-) plans have different issue about dots, we should fix that too, no bug for that one, @qiyuann do you think we need a bug for the plans and migrations ?

not sure I'm following - that was just an example, one can replace plans with providers or mappings or any other entity that doesn't work the same way in the CLI and UI

qiyuann:
> EDIT @ahadas i copied your remark, it is providers not plans :-) plans have different issue about dots, we should fix that too, no bug for that one, @qiyuann do you think we need a bug for the plans and migrations ?

@yaacov Filed a bug for plans and maps, https://github.com/kubev2v/forklift-console-plugin/issues/714

yaacov:
> but if there's a proper message in the UI for that as @qiyuann noted above, then we can skip it

@ahadas yes, my point is that I think we can skip the note in the providers too (it can be documented as a known bug) :-)

As a rule of thumb, messages are not read by users, as a user if I want to make a plan named "plan.8" I will not care what the help message say, it's a bug for me.

IMHO the fact that you can't create plan, provider or mappings in the UI as you can in the CLI is a bug, and it should be documented as "known bug" not as a feature, regardless of help messages.

RichardHoch:
@ahadas @yaacov @qiyuann @sgratch  Please see my note in mtv-qe about this.

documentation/modules/adding-source-provider.adoc
```diff
@@ -26,37 +26,176 @@ You can add {a-rhv} source provider by using the {ocp} web console.
...
 
-* *Provider name*: Name to display in the list of providers
-* *vCenter host name or IP address*: vCenter host name or IP address - if a certificate for FQDN is specified, the value of this field needs to match the FQDN in the certificate
-* *vCenter user name*: vCenter user, for example, `user@vsphere.local`
-* *vCenter password*: vCenter user password
-* *VDDK init image*: VDDKInitImage path
+* *Provider resource name*: Name to display in the list of providers.
++
+include::snip-provider-names.adoc[]
+* *URL*: Endpoint.
```
yaacov, line 79: can we have more information about what a vsphere endpoint is, maybe where I take it from ?

I don't know the answer to the above question :-) @ahadas can you point @RichardHoch to the relevant information ?

yaacov: "vCenter host name or IP address - if a certificate for FQDN is specified, the value of this field needs to match the FQDN in the certificate" ?

documentation/modules/migrating-virtual-machines-cli.adoc
```diff
@@ -80,22 +80,24 @@ $ cat << EOF | {oc} apply -f -
...
+`.`
+<2> Allowed values are `ovirt`, `vsphere`, and `openstack`.
```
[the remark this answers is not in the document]
yaacov: and `openshift`

yaacov:
LGTM
( I'm ok with adding a comment about providers names )

documentation/modules/migrating-virtual-machines-cli.adoc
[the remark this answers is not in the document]
yaacov: I see it's changed in a different PR, resolving

RichardHoch:
@ahadas @yaacov @sgratch @qiyuann 
Revised after our discussion in Slack on Wednesday afternoon (Israel time). The PR now includes only the note that Provider names cannot include a (.). I formatted the note according to Red Hot documentation standards.

documentation/modules/adding-source-provider.adoc
yaacov:
```suggestion
* *URL*: vCenter host name or IP address - if a certificate for FQDN is specified, the value of this field needs to match the FQDN in the certificate.
```

documentation/modules/snip-provider-names.adoc
```diff
@@ -0,0 +1,7 @@
+:_content-type: SNIPPET
+
+[NOTE]
+====
+Provider names cannot include the following character: +
+`.`
+====
```
sgratch, line 7: You missed starting with an alphabet char limitation.
Meaning that it should be something like:
"Provider names **must start with an alphabet char** and cannot include the following character: `.`"

(the provider names validation is a bit different than mapping and plans since it blocks starting with a alphanumeric char as well)

yaacov:
@RichardHoch can you fix the help text for vsphere URL, see https://github.com/kubev2v/forklift-documentation/pull/419#discussion_r1317202105

sgratch (commented):
Please check my comment

RichardHoch:
@yaacov I'll make the change you asked for in the Create Providers PR. Let's keep this PR focused just on the one issue.

documentation/modules/snip-provider-names.adoc
```diff
@@ -0,0 +1,7 @@
+:_content-type: SNIPPET
+
+[NOTE]
+====
+Provider names musi start with an alaphabetic character and cannot include the following character: +
```
sgratch, line 5: A typo musi -> must 
```suggestion
Provider names must start with an alaphabetic character and cannot include the following character: +
```

documentation/modules/adding-source-provider.adoc
```diff
@@ -26,37 +26,176 @@ You can add {a-rhv} source provider by using the {ocp} web console.
...
+* *Provider resource name*: Name to display in the list of providers
++
+include::snip-provider-names.adoc[]
+* *URL*: URL of the NFS file share that serves the OVA
+endif::[]
+ifdef::ostack[]
+. Click *OpenStack*.
+. Specify the following fields:
+
+* *Provider resource name*: Name to display in the list of providers.
```
sgratch, line 129: Missing adding the note of `snip-provider-names` for OpenStack as well

```suggestion
* *Provider resource name*: Name to display in the list of providers.
+
 include::snip-provider-names.adoc[]
```

documentation/modules/adding-source-provider.adoc
yaacov: @RichardHoch thanks, I see it in the other PR

RichardHoch:
@sgratch Made the changes you suggested. Please review again.

sgratch (approved):
LGTM

qiyuann (approved):
LGTM

ahadas:
> IMHO the fact that you can't create plan, provider or mappings in the UI as you can in the CLI is a bug,

+1
that's why we plan to resolve it

> and it should be documented as "known bug" not as a feature, regardless of help messages.

@yaacov that's where we see things differently, when it comes to the meaning of a note - in my view, having a note doesn't necessarily makes what's inside a feature, in this case it's a documented limitation

ahadas (approved):
need to be careful with the rebase here

RichardHoch:
> need to be careful with the rebase here

@ahadas I based this PR on the Create Providers PR, instead of on "main", to avoid rebase problems. But thanks for keeping me honest (and avoiding me screaming my head off at 2 AM Thursday night...)

documentation/modules/adding-source-provider.adoc
```diff
@@ -26,37 +26,178 @@ You can add {a-rhv} source provider by using the {ocp} web console.
...
+====
+endif::[]
+ifdef::ostack[]
+= Adding an {osp} source provider
+
+You can add an {osp} source provider by using the {ocp} web console.
+
+[NOTE]
+====
+Migration using {osp} source providers only supports VMs that use only Cinder volumes.
```
CarmiWisemon, line 49:
```suggestion
Migration using {osp} source providers supports VMs that use only Cinder volumes.
```

documentation/modules/adding-source-provider.adoc
```diff
@@ -26,37 +26,178 @@ You can add {a-rhv} source provider by using the {ocp} web console.
...
+ifdef::cnv[]
+= Adding a Red Hat {virt} source provider
+
+You can add a Red Hat {virt} source provider by using the {ocp} web console.
+
+endif::[]
+ifdef::cnv2[]
+= Adding {a-virt} destination provider
+
+You can add another {virt} destination provider to the {ocp} web console in addition to the default {virt} destination provider, which is the provider where you installed {project-short}.
```
CarmiWisemon, line 61:
```suggestion
You can add another {virt} destination provider to the {ocp} web console, in addition to the default {virt} destination provider, which is the provider where you installed {project-short}.
```

documentation/modules/snip-provider-names.adoc
```diff
@@ -0,0 +1,7 @@
+:_content-type: SNIPPET
+
+[NOTE]
+====
+Provider names must start with an alaphabetic character and cannot include the following character: +
```
CarmiWisemon, line 5:
```suggestion
Provider names must start with an alphabetic character and cannot include the following character: +
```

documentation/modules/adding-source-provider.adoc
```diff
@@ -26,37 +26,178 @@ You can add {a-rhv} source provider by using the {ocp} web console.
...
++
+include::snip-provider-names.adoc[]
+* *URL*: Endpoint of the API server
+* *Service account bearer token*: User or service account bearer token
+endif::[]
+ifdef::cnv2[]
+. Click *{virt}*.
+. Specify the following fields:
+
+* Provider resource name*: Name to display in the list of providers
```
CarmiWisemon, line 186:
```suggestion
*Provider resource name*: Name to display in the list of providers
```

documentation/modules/adding-source-provider.adoc
RichardHoch: Changes the meaning, making the statement less restrictive.

documentation/modules/snip-provider-names.adoc
RichardHoch: Thanks.

CarmiWisemon:
LGTM

documentation/modules/adding-source-provider.adoc
anarnold97: Yeah ""only ... only** looks clumsy, however it actually serves to emphasize the restrictions

documentation/modules/adding-source-provider.adoc
```diff
@@ -26,37 +26,178 @@ You can add {a-rhv} source provider by using the {ocp} web console.
...
 . Click *Create* to add and save the provider.
 +
-The source provider appears in the list of providers.
+The provider appears in the list of providers.
+
+ifdef::ova[]
++
+[NOTE]
+====
+A error message might appear that states that an error has occured. You can ignore this message.
```
anarnold97, line 201:
```suggestion
An error message might appear that states that an error has occured. You can ignore this message.
```

anarnold97 (approved):

PR #426, opened 2023-09-20
MTV-271: RHEV_Permissions

sbeskin-redhat:
Resolves: https://issues.redhat.com/browse/MTV-271
MTV 2.5.1
Preview:

### DIFF
--- a/documentation/modules/rhv-prerequisites.adoc
+++ b/documentation/modules/rhv-prerequisites.adoc
@@ -8,7 +8,7 @@
 
 The following prerequisites apply to {rhv-full} migrations:
 
-* You must log in as a user with at least `DiskCreator` and `UserVmManager` privileges.
+* You must have {rhv-short} admin permissions to be able to migrate any virtual machines. Alternatively, you need to have at least the `DiskCreator` and `UserVmManager` permissions granted by an `admin` user per each individual virtual machine you want to migrate.
 * You must use a xref:compatibility-guidelines_{context}[compatible version] of {rhv-full}.
 * You must have the {manager} CA certificate, unless it was replaced by a third-party certificate, in which case, specify the {manager} Apache CA certificate.
 +
--- a/documentation/modules/rhv-prerequisites.adoc
+++ b/documentation/modules/rhv-prerequisites.adoc
@@ -9,8 +9,8 @@
 The following prerequisites apply to {rhv-full} migrations:
 
 * You must have one of the following:
-** {rhv-short} admin permissions. These permissions allows you to migrate any virtual machine in the system.
-** `DiskCreator` and `UserVmManager` permissions: These permissions have to be granted by an `admin` user to you individually per each individual virtual machine that you want to migrate.
+** {rhv-short} admin permissions. These permissions allow you to migrate any virtual machine in the system.
+** `DiskCreator` and `UserVmManager` permissions: These permissions have to be granted to you individually by an `admin` user per each individual virtual machine that you want to migrate.
 * You must use a xref:compatibility-guidelines_{context}[compatible version] of {rhv-full}.
 * You must have the {manager} CA certificate, unless it was replaced by a third-party certificate, in which case, specify the {manager} Apache CA certificate.
 +
--- a/documentation/modules/rhv-prerequisites.adoc
+++ b/documentation/modules/rhv-prerequisites.adoc
@@ -10,7 +10,7 @@ The following prerequisites apply to {rhv-full} migrations:
 
 * You must have one of the following:
 ** {rhv-short} admin permissions. These permissions allow you to migrate any virtual machine in the system.
-** `DiskCreator` and `UserVmManager` permissions: These permissions have to be granted to you individually by an `admin` user per each individual virtual machine that you want to migrate.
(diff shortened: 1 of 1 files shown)


documentation/modules/rhv-prerequisites.adoc
```diff
@@ -8,7 +8,7 @@
 
 The following prerequisites apply to {rhv-full} migrations:
 
-* You must have {rhv-short} admin permissions to be able to migrate any virtual machines. Alternatively, you need to have at least the `DiskCreator` and `UserVmManager` permissions granted by an `admin` user per each individual virtual machine you want to migrate.
+* You must have the {rhv-short} admin permissions to be able to migrate any virtual machines. Alternatively, you need to have at least the `DiskCreator` and `UserVmManager` permissions granted by an `admin` user per each individual virtual machine you want to migrate.
```
bkhizgiy, line 11: looking better now, thanks.
few notes, I would emphasize that the permissions that needs to be granted are for the specific user that will perform the migration, something like:
`Alternatively, you need to have at least the `DiskCreator` and `UserVmManager` permissions set for the desired user per each individual virtual machine you want to migrate, this can be done by the admin user.`

Also I would drop the `you must have` part, and just display the two existing options, something like:
In order to migrate a virtual machine you may use admin account, which will give you access to  any virtual machine in the system. Alternatively ...`

sbeskin-redhat: @bkhizgiy 
Please take a look now.

bkhizgiy (approved):
Thanks, looking better now approved by me.

documentation/modules/rhv-prerequisites.adoc
```diff
@@ -8,6 +8,9 @@
 
 The following prerequisites apply to {rhv-full} migrations:
 
+* You must have one of the following:
+** {rhv-short} admin permissions. These permissions allow you to migrate any virtual machine in the system.
+** `DiskCreator` and `UserVmManager` permissions: These permissions have to be granted to you individually by an `admin` user per each individual virtual machine that you want to migrate.
```
RichardHoch, line 13: Suggestions:
1. End "DiskCreator` and `UserVmManager` permissions:" with a period as you did with RHV admin permissions. Alternatively, use a colon after "permissions" in both cases, although I think a period is better. 
2. I don't think you need the word "individually" in "granted to you individually, " unless you want to say it can't be granted by a role-based access control (RBAC) permission set.
3. You probably don't need "individual"  in "each individual VM" either, although in the latter case, it does emphasize that permission is granted per machine, and nit for a group of them, so I'd say there's a good case for leaving it.

sbeskin-redhat: @RichardHoch 
1. Fixed, thank you
2, 3. Yes, that's the point of the option - the permissions should be granted to an individual user, not to a role, by an admin and specifically for each virtual machine. So I think I have to keep 'individually'. Are you OK with that?

RichardHoch: OK.

RichardHoch: LGTM

bkhizgiy:
@qiyuann could you please take a look as well?

qiyuann (commented):
I created a RHV user with `DiskCreator` and `UserVmManager` permissions, but the RHV provider using this user was always on staging status. 
Need to investigate what should be the correct permissions.

sbeskin-redhat:
@qiyuann 
Thank you! Please let me know when you have results.

RichardHoch:
@bkhizgiy @qiyuann Hi! This item has been assigned to me -- can you please review this PR again?

bkhizgiy:
@RichardHoch sorry, missed the comments:) it seams OK to me, but let me do one more testing tomorrow, it's been a while.. I"ll update the status here after that.

documentation/modules/rhv-prerequisites.adoc
```diff
@@ -8,6 +8,9 @@
 
 The following prerequisites apply to {rhv-full} migrations:
 
+* You must have one of the following:
+** {rhv-short} admin permissions. These permissions allow you to migrate any virtual machine in the system.
+** `DiskCreator` and `UserVmManager` permissions. These permissions have to be granted to you individually by an `admin` user per each individual virtual machine that you want to migrate.
```
bkhizgiy, line 13: In addition to the individual permissions for VMs, any user who wants to add a new provider to the system must have at least the UserRole and ReadOnlyAdmin roles assigned to them. These are the minimum required permissions, however, any other administrative or superuser permissions will also work.

bkhizgiy:
@RichardHoch, to summarize, these are the required permissions:

In general, for each user to be able to add a provider, they must have the `UserRole` and at least `ReadOnlyAdmin`.
For migrating each individual VM, the user must have `UserVmManager` and `DiskCreator` permissions.

documentation/modules/rhv-prerequisites.adoc
RichardHoch: @bkhizgiy Are you saying you need these roles in addition to `DiskCreator` and `UserVmManager` permissions or instead of them?

RichardHoch:
@bkhizgiy @ahadas  I opened a new PR, https://github.com/kubev2v/forklift-documentation/pull/491, so I could work on the Jira.

PR #427, opened 2023-09-21
Updating VMWare pre-requisites to include important note

anarnold97:
@ahadas - rather than a note, I thought it would be best to include at the very top of this section. 
I hope this is ok. 
image


anarnold97:
Arik approved in Slack

PR #432, opened 2023-10-03
Align description of provider properties with the UI

RichardHoch:
MTV 2.5.1

Resolves: https://issues.redhat.com/browse/MTV-717 and https://issues.redhat.com/browse/MTV-703 by:
1. Changing the description of the URL field for vSphere source providers.
2. Changing the description of the URL field for RHV source providers.
3. Removing note that said provider names cannot include "."
4-6: Removing an inconsistent "/" in some URLs. 

Previews:
1. https://file.emea.redhat.com/rhoch/251_provider_fixes/html-single/#adding-source-providers [See definitions of "URL"  for vSphere and RHV source providers; also see that note about provider names was removed]
2. https://file.emea.redhat.com/rhoch/251_provider_fixes/html-single/#migrating-virtual-machines-cli_mtv [Step 1, callout 11 and steep 2, callout 2] 
3. https://file.emea.redhat.com/rhoch/251_provider_fixes/html-single/#migrating-virtual-machines-cli_mtv [step 6, callout 2]

### DIFF
--- a/documentation/modules/adding-source-provider.adoc
+++ b/documentation/modules/adding-source-provider.adoc
@@ -78,9 +78,9 @@ ifdef::vmware[]
 . Specify the following fields:
 
 * *Provider resource name*: Name of the source provider.
-* *URL*: vCenter host name or IP address - if a certificate for FQDN is specified, the value of this field needs to match the FQDN in the certificate.
+* *URL*: URL of the SDK endpoint of the vCenter on which the source VM is mounted. Ensure that the URL includes the `sdk` path, usually `/sdk`. For example, `https://vCenter-host-example.com/sdk`. If a certificate for FQDN is specified, the value of this field needs to match the FQDN in the certificate.
 * *VDDK init image*: `VDDKInitImage` path. It is strongly recommended to create a VDDK init image to accelerate migrations. For more information, see xref:../master.adoc#creating-vddk-image_mtv[Creating a VDDK image].
-* *Username*: vCenter user, for example, `user@vsphere.local`.
+* *Username*: vCenter user. For example, `user@vsphere.local`.
 * *Password*: vCenter user password.
 * *SHA-1 fingerprint*: The provider currently requires the SHA-1 fingerprint of the vCenter Server's TLS certificate in all circumstances. vSphere calls this the server's thumbprint.
 +
@@ -97,12 +97,10 @@ ifdef::rhv[]
 . Click *Red Hat Virtualization*
 . Specify the following fields:
 
-* *Provider resource name*: Name of the source provider
-+
-include::snip-provider-names.adoc[]
-* *URL*: API endpoint of the {rhv-short} Manager on which the source VM is mounted
-* *Username*: Username
-* *Password*: Password
+* *Provider resource name*: Name of the source provider.
+* *URL*: URL of the API endpoint of the {rhv-full} Manager (RHVM) on which the source VM is mounted. Ensure that the URL includes the path leading to the RHVM API server, usually `/ovirt-engine/api`. For example, `https://rhv-host-example.com/ovirt-engine/api`.
+* *Username*: Username.
+* *Password*: Password.
 
 . Choose one of the following options for validating CA certificates:
 
@@ -117,8 +115,6 @@ ifdef::ova[]
 . Specify the following fields:
 
 * *Provider resource name*: Name of the source provider
-+
-include::snip-provider-names.adoc[]
 * *URL*: URL of the NFS file share that serves the OVA
 endif::[]
 ifdef::ostack[]
@@ -126,9 +122,7 @@ ifdef::ostack[]
 . Specify the following fields:
 
 * *Provider resource name*: Name of the source provider.
-+
-include::snip-provider-names.adoc[]
-* *URL*: {osp} Identity (Keystone) endpoint, for example, `http://controller:5000/v3`.
+* *URL*: URL of the {osp} Identity (Keystone) endpoint. For example, `http://controller:5000/v3`.
 * *Authentication type*: Choose one of the following methods of authentication and supply the information related to your choice. For example, if you choose *Application credential ID* as the authentication type, the *Application credential ID* and the *Application credential secret* fields become active, and you need to supply the ID and the secret.
 
 ** *Application credential ID*
@@ -171,9 +165,7 @@ ifdef::cnv[]
 . Specify the following fields:
 
 * *Provider resource name*: Name of the source provider
-+
-include::snip-provider-names.adoc[]
-* *URL*: Endpoint of the API server
+* *URL*: URL of the endpoint of the API server
 * *Service account bearer token*: Token for a service account with `cluster-admin` privileges
 +
 If both *URL* and *Service account bearer token* are left blank, the local {ocp-short} cluster is used.
@@ -183,9 +175,7 @@ ifdef::cnv2[]
 . Specify the following fields:
 
 * *Provider resource name*: Name of the source provider
-+
-include::snip-provider-names.adoc[]
-* *URL*: Endpoint of the API server
+* *URL*: URL of the endpoint of the API server
 * *Service account bearer token*: Token for a service account with `cluster-admin` privileges
 +
 If both *URL* and *Service account bearer token* are left blank, the local {ocp-short} cluster is used.
--- a/documentation/modules/creating-validation-rule.adoc
+++ b/documentation/modules/creating-validation-rule.adoc
@@ -121,7 +121,7 @@ spec:
 EOF
 ----
 <1> Allowed values are `ovirt`, `vsphere`, and `openstack`.
-<2> Specify the API end point URL, for example, `https://<vCenter_host>/sdk` for vSphere, `https://<engine_host>/ovirt-engine/api/` for {rhv-short}, or `https://<identity_service>/v3` for {osp}.
+<2> Specify the API end point URL, for example, `https://<vCenter_host>/sdk` for vSphere, `https://<engine_host>/ovirt-engine/api` for {rhv-short}, or `https://<identity_service>/v3` for {osp}.
 <3> Specify the name of the provider `Secret` CR.
 
 You must update the rules version after creating a custom rule so that the `Inventory` service detects the changes and validates the VMs.
--- a/documentation/modules/migrating-virtual-machines-cli.adoc
+++ b/documentation/modules/migrating-virtual-machines-cli.adoc
@@ -73,7 +73,7 @@ EOF
 <8> {osp} only: Specify the project name.
 <9> {osp} only: Specify the name of the {osp} region.
 <10> {rhv-short} and {osp} only: For {rhv-short}, enter the {manager} CA certificate unless it was replaced by a third-party certificate, in which case enter the {manager} Apache CA certificate. You can retrieve the {manager} CA certificate at https://<engine_host>/ovirt-engine/services/pki-resource?resource=ca-certificate&format=X509-PEM-CA. For {osp}, enter the CA certificate for connecting to the source environment. The certificate is not used when `insecureSkipVerify` is set to `<true>`.
-<11> Specify the API end point URL, for example, `https://<vCenter_host>/sdk` for vSphere, `https://<engine_host>/ovirt-engine/api/` for {rhv-short}, or `https://<identity_service>/v3` for {osp}.
+<11> Specify the API end point URL, for example, `https://<vCenter_host>/sdk` for vSphere, `https://<engine_host>/ovirt-engine/api` for {rhv-short}, or `https://<identity_service>/v3` for {osp}.
 <12> VMware only: Specify the vCenter SHA-1 fingerprint.
 +
 [NOTE]
@@ -110,7 +110,7 @@ spec:
 EOF
 ----
 <1> Allowed values are `ovirt`, `vsphere`, and `openstack`.
-<2> Specify the API end point URL, for example, `https://<vCenter_host>/sdk` for vSphere, `https://<engine_host>/ovirt-engine/api/` for {rhv-short}, or `https://<identity_service>/v3` for {osp}.
+<2> Specify the API end point URL, for example, `https://<vCenter_host>/sdk` for vSphere, `https://<engine_host>/ovirt-engine/api` for {rhv-short}, or `https://<identity_service>/v3` for {osp}.
 <3> VMware only: Specify the VDDK image that you created.
 <4> Specify the name of provider `Secret` CR.
 
--- a/documentation/modules/snip-provider-names.adoc
+++ b/documentation/modules/snip-provider-names.adoc
@@ -1,7 +1,0 @@
-:_content-type: SNIPPET
-
-[NOTE]
-====
-Provider names must start with an alphabetic character and cannot include the following character: +
-`.`
-====


documentation/modules/adding-source-provider.adoc
```diff
@@ -100,7 +100,7 @@ ifdef::rhv[]
 * *Provider resource name*: Name of the source provider
 +
 include::snip-provider-names.adoc[]
-* *URL*: API endpoint of the {rhv-short} Manager on which the source VM is mounted
+* *URL*: API endpoint of the {rhv-full} ({rhv-short}) engine on which the source VM is mounted, for example, `https://rhv-host-example.com/ovirt-engine/api`
```
ahadas, line 103: as I commented on #744 - better to keep RHVM here rather than using "engine"

RichardHoch:
@bennyz @bkhizgiy @sgratch @yaacov @ahadas  Please review this PR. It focuses on the changes we need to make to the documentation for the Create Providers UI (https://access.redhat.com/documentation/en-us/migration_toolkit_for_virtualization/2.5/html-single/installing_and_using_the_migration_toolkit_for_virtualization/index#adding-providers) -- NOT the error messages, only the documentation of the fields. The error messages are in other PRs.

documentation/modules/adding-source-provider.adoc
```diff
@@ -78,7 +78,7 @@ ifdef::vmware[]
 . Specify the following fields:
 
 * *Provider resource name*: Name of the source provider.
-* *URL*: vCenter host name or IP address - if a certificate for FQDN is specified, the value of this field needs to match the FQDN in the certificate.
+* *URL*: API endpoint of the vCenter on which the source VM is mounted, for example, `https://vCenter-host-example.com/sdk`. If a certificate for FQDN is specified, the value of this field needs to match the FQDN in the certificate.
```
ahadas, line 81: as I commented on #745, better to use `vCenter SDK endpoint`

yaacov:
@RichardHoch thank you for opening this PR :green_heart:

I'll ping here once the text in https://github.com/kubev2v/forklift-console-plugin/pull/744 and https://github.com/kubev2v/forklift-console-plugin/pull/745 are merged so we have a stable point of reference.

documentation/modules/adding-source-provider.adoc
RichardHoch: Fixed.

RichardHoch:
@ahadas I made the changes you suggested. Please review the PR again.

documentation/modules/adding-source-provider.adoc
```diff
@@ -100,7 +100,7 @@ ifdef::rhv[]
 * *Provider resource name*: Name of the source provider
 +
 include::snip-provider-names.adoc[]
-* *URL*: API endpoint of the {rhv-short} Manager on which the source VM is mounted
+* *URL*: API endpoint of the {rhv-full} ({rhv-short}) Manager on which the source VM is mounted, for example, `https://rhv-host-example.com/ovirt-engine/api`
```
sgratch, line 103: It seems that currently it appears as:
"API endpoint of the **Red Hat Virtualization (RHV) Manager**...."

 It's ok, but to be aligned with the UI, it's better to change to:

"API endpoint of the **Red Hat Virtualization Manager (RHVM)** ...."

```suggestion
 *URL*: API endpoint of the {rhv-full} Manager  ({rhv-short})M on which the source VM is mounted, for example, `https://rhv-host-example.com/ovirt-engine/api`
```

sgratch (commented):
LGTM, except one comment.

documentation/modules/adding-source-provider.adoc
RichardHoch: Fxed.

RichardHoch:
@sgratch Made the change you suggested.

documentation/modules/adding-source-provider.adoc
```diff
@@ -100,7 +100,7 @@ ifdef::rhv[]
 * *Provider resource name*: Name of the source provider
 +
 include::snip-provider-names.adoc[]
-* *URL*: API endpoint of the {rhv-short} Manager on which the source VM is mounted
+* *URL*: API endpoint of the {rhv-full} Manager (RHVM) on which the source VM is mounted, for example, `https://rhv-host-example.com/ovirt-engine/api`
```
yaacov, line 103: in https://github.com/kubev2v/forklift-console-plugin/pull/744/files#diff-075b2af748cc5c8407a5cd156371011523517ef954d89d2bac7d9d5c72d9fe97R29

we chose to write:

> URL of the Red Hat Virtualization Manager (RHVM) API endpoint. Ensure the URL includes the "/ovirt-engine/api" path. for example: https://rhv-host-example.com/ovirt-engine/api

I'm ok with current version and ok with uipdateing to match to UI

documentation/modules/adding-source-provider.adoc
```diff
@@ -78,7 +78,7 @@ ifdef::vmware[]
 . Specify the following fields:
 
 * *Provider resource name*: Name of the source provider.
-* *URL*: vCenter host name or IP address - if a certificate for FQDN is specified, the value of this field needs to match the FQDN in the certificate.
+* *URL*: SDK endpoint of the vCenter on which the source VM is mounted, for example, `https://vCenter-host-example.com/sdk`. If a certificate for FQDN is specified, the value of this field needs to match the FQDN in the certificate.
```
yaacov, line 81: In https://github.com/kubev2v/forklift-console-plugin/pull/745/files#diff-78ab8802dcf0222dfbd15eeeac4d1f58cefaec2614e5900cbd2146cd8738a9d8R30

we chose to wirte:

> URL of the vCenter SDK endpoint. Ensure the URL includes the "/sdk" path. For example: https://vCenter-host-example.com/sdk

I'm ok with current version and ok with uipdateing to match to UI

yaacov: p.s. 

i see the fqdn comment apear here:
 If a certificate for FQDN is specified, the value of this field needs to match the FQDN in the certificate.
 
@ahadas  if it's needed we can add it to the UI, and leave it here ?

documentation/modules/adding-source-provider.adoc
yaacov: p.s. not that the `for` -> `For` was fixed in the UI in a folowup

yaacov (approved):
LGTM

@RichardHoch  hi, not exactly as the UI, but the UI does not need to be verbatim copy of the docs (*)

I commented with the texts chosen for the UI copy of the docs, we can also do a PR for the UI to alight to the texts we end up using here.

cc:// @sgratch 

(*) note that the docs is the "source of truth" for the UI, so the UI need to match the docs not the other way  around.

documentation/modules/adding-source-provider.adoc
yaacov: p.p.s 
from the text in the UI it sounds like it must end with /sdk ... (same with the oVIrt texts)
the URL does not need to end with this post fixes, it's a little misleading, but IMHO we can trust our users to understand this nuance

RichardHoch: @yaacov @sgratch "Ensure" means it must be so. If that is not 100% accurate, that is, if it is OK if the URL does not  end in /sdk, then you might want do one of the following in the UI:

- Remove the sentence.
- Change it to "Ensure.../sdk path, if needed."
- Change it to "It is recommended that URL includes..."  [We try to avoid saying "it is recommended" but if you think it is useful, it is OK]

I'm not sure which of the above is technically correct. You can leave the UI alone, but I'd recommend you change it if the text might confuse users.

documentation/modules/adding-source-provider.adoc
RichardHoch: @yaacov I added "URL of the..." to all the URL descriptions.

documentation/modules/adding-source-provider.adoc
yaacov: thanks @RichardHoch , it did sound off to me.
IMHO "if needed" is best here, we want to ensure the user understand that the URL includes the postifx that can be "/sdk" but can also be other strings like "/east1/sdk/v1" or "/server",

RichardHoch:
@ahadas @yaacov Added "URL of the.." to any URL definitions that did not have it. Please review again. I left the "Ensure ..." sentence out of the documentation since it sounded like it was not 100% necessary.

documentation/modules/adding-source-provider.adoc
yaacov: @RichardHoch before this change we used the term "sdk path" to denote any path that lead to the sdk, instead of using "/sdk" that sounds like it must be "/sdk".
Can we say that it needs to include the sdk path, and then say that "/sdk" is the common path used as the sdk path ?


https://github.com/kubev2v/forklift-console-plugin/pull/745/files#diff-78ab8802dcf0222dfbd15eeeac4d1f58cefaec2614e5900cbd2146cd8738a9d8L111

RichardHoch:
@yaacov Our comments are overlapping. Would It be better if I add "Ensure..., if needed." to the documentation?

yaacov:
> @yaacov Our comments overlapped. Would it be better if I add "Ensure..., if needed." to the documentation?

makes sens to me, @ahadas WDYT ?

p.s. this comment is also for the ovirt engine path, it can be "ovirt-engine/api" and can also be some other pass leading to the ovirt engine api.

yaacov:
@RichardHoch the reason we changed the UI text to include the "please make sure you add the ovirt engine path in oVirt and the sdk path vSphere" was because of users feedback, in your current version, i don't see this emphasis on the need to add the API path.
@ahadas we can drop the "Ensure you add "/skd" "/ovier-engine" " but then, we will return to the original phrasing, the one that users didn't like, because they tried to enter the URL without the path, WDYT ?

yaacov:
@RichardHoch @sgratch I see the discussion moved here :-)
This is good, beacuse the docs is the "source of truth" for the UI.

@sgratch once we agree on text here, please make a PR to copy it the UI.

RichardHoch:
@yaacov @ahadas How's this?

yaacov (approved):
LGTM

sgratch:
- Regarding the FQDN comment appears here and not in the UI:
"`If a certificate for FQDN is specified, the value of this field needs to match the FQDN in the certificate."
`
  I intentionally skipped that part in the UI since it seems to me like too much information to put for a field help text. And anyway,  we can consider adding this as a help text for the `CA certificate` field and not for the `URL` field.

- Regarding writing "`Ensure the URL includes the xxx path...."` for both  RHV and vSphere providers URLs:
 Another option in addition to Richard's suggestions is to to replace it with something like:
 "`usually the URL includes the xxx path. For example..."`

documentation/modules/adding-source-provider.adoc
```diff
@@ -78,9 +78,9 @@ ifdef::vmware[]
 . Specify the following fields:
 
 * *Provider resource name*: Name of the source provider.
-* *URL*: vCenter host name or IP address - if a certificate for FQDN is specified, the value of this field needs to match the FQDN in the certificate.
+* *URL*: URL of the SDK endpoint of the vCenter on which the source VM is mounted. Ensure that the URL includes the `sdk` path, usually `.../sdk`. For example, `https://vCenter-host-example.com/sdk`. If a certificate for FQDN is specified, the value of this field needs to match the FQDN in the certificate.
```
ahadas, line 81:
```suggestion
* *URL*: URL of the SDK endpoint of the vCenter on which the source VM is mounted. Ensure that the URL includes the `sdk` path, usually `/sdk`. For example, `https://vCenter-host-example.com/sdk`. If a certificate for FQDN is specified, the value of this field needs to match the FQDN in the certificate.
```

documentation/modules/adding-source-provider.adoc
```diff
@@ -97,12 +97,12 @@ ifdef::rhv[]
...
 . Specify the following fields:
 
-* *Provider resource name*: Name of the source provider
+* *Provider resource name*: Name of the source provider.
 +
 include::snip-provider-names.adoc[]
-* *URL*: API endpoint of the {rhv-short} Manager on which the source VM is mounted
-* *Username*: Username
-* *Password*: Password
+* *URL*: URL of the API endpoint of the {rhv-full} Manager (RHVM) on which the source VM is mounted. Ensure that the URL includes the path leading to the ovirt engine API, usually `.../ovirt-engine/api` For example, `https://rhv-host-example.com/ovirt-engine/api`.
```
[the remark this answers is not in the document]
ahadas:
```suggestion
* *URL*: URL of the API endpoint of the {rhv-full} Manager (RHVM) on which the source VM is mounted. Ensure that the URL includes the path leading to the RHVM API server, usually `/ovirt-engine/api`. For example, `https://rhv-host-example.com/ovirt-engine/api`.
```

ahadas:
> * Regarding the FQDN comment appears here and not in the UI:
>   "`If a certificate for FQDN is specified, the value of this field needs to match the FQDN in the certificate." `
>   I intentionally skipped that part in the UI since it seems to me like too much information to put for a field help text. And anyway,  we can consider adding this as a help text for the `CA certificate` field and not for the `URL` field.

yeah, I don't even think we must add that to the UI - from the cases we get, it looks like insecure connections become more common (but it's good to have it in the documentation and I think the way it is now, is better than moving it to the CA certificate field)

> * Regarding writing "`Ensure the URL includes the xxx path...."` for both  RHV and vSphere providers URLs:
>   Another option ,in addition to Richard's suggestions, is to to replace it with something like:
>   "`usually the URL includes the xxx path. For example..."`
>   The "...,if needed" suggested option seems a bit confusing to me, because the user won't know what are the terms for "needed".

I guess you referenced a previous state of this PR as I don't see `... , if needed` now

sgratch (approved):
LGTM

RichardHoch:
@ahadas @yaacov @sgratch Please review again.

documentation/modules/adding-source-provider.adoc
```diff
@@ -78,9 +78,9 @@ ifdef::vmware[]
 . Specify the following fields:
 
 * *Provider resource name*: Name of the source provider.
-* *URL*: vCenter host name or IP address - if a certificate for FQDN is specified, the value of this field needs to match the FQDN in the certificate.
+* *URL*: URL of the SDK endpoint of the vCenter on which the source VM is mounted. Ensure that the URL includes the `sdk` path, usually `sdk`. For example, `https://vCenter-host-example.com/sdk`. If a certificate for FQDN is specified, the value of this field needs to match the FQDN in the certificate.
```
ahadas, line 81:
```suggestion
* *URL*: URL of the SDK endpoint of the vCenter on which the source VM is mounted. Ensure that the URL includes the `sdk` path, usually `/sdk`. For example, `https://vCenter-host-example.com/sdk`. If a certificate for FQDN is specified, the value of this field needs to match the FQDN in the certificate.
```

documentation/modules/adding-source-provider.adoc
```diff
@@ -97,12 +97,10 @@ ifdef::rhv[]
...
 . Specify the following fields:
 
-* *Provider resource name*: Name of the source provider
-+
-include::snip-provider-names.adoc[]
-* *URL*: API endpoint of the {rhv-short} Manager on which the source VM is mounted
-* *Username*: Username
-* *Password*: Password
+* *Provider resource name*: Name of the source provider.
+* *URL*: URL of the API endpoint of the {rhv-full} Manager (RHVM) on which the source VM is mounted. Ensure that the URL includes the path leading to the RHVM API server, usually `ovirt-engine/api`. For example, `https://rhv-host-example.com/ovirt-engine/api`.
```
ahadas, line 101:
```suggestion
* *URL*: URL of the API endpoint of the {rhv-full} Manager (RHVM) on which the source VM is mounted. Ensure that the URL includes the path leading to the RHVM API server, usually `/ovirt-engine/api`. For example, `https://rhv-host-example.com/ovirt-engine/api`.
```

ahadas (commented):
minor comment about the paths to align them with the message in the UI
and @RichardHoch we can we also remove the file snip-provider-names.adoc, right?

yaacov (approved):
lgtm

p.s. 
you changed the OCP provider text, `URL of the endpoint of the API server`
we will need to add more text here, but I do not think we need to change the text here, we will  go over all the texts in the provider new creation form in Issue https://github.com/kubev2v/forklift-console-plugin/issues/646, @sgratch

sgratch:
> p.s. you changed the OCP provider text, `URL of the endpoint of the API server` we will need to add more text here, but I do not think we need to change the text here, we will go over all the texts in the provider new creation form in Issue [kubev2v/forklift-console-plugin#646](https://github.com/kubev2v/forklift-console-plugin/issues/646), @sgratch

It's ok to leave it just to have all URL text msgs start with: "URL of the....".
It will be revised with #646 of course, but for now it's better to leave it as it is now.

ahadas (approved):

RichardHoch:
@qiyuann This is ready for you.

qiyuann:
LGTM

qiyuann (approved):

anarnold97 (approved):
@RichardHoch - good work

PR #474, opened 2024-02-26
Add vSphere endpoint options

RichardHoch:
MTV 2.6

Resolves https://issues.redhat.com/browse/MTV-942 and https://issues.redhat.com/browse/MTV-553 by adding vSphere endpoint options to the procedure for creating a vSphere source provider using the UI.

### DIFF
--- a/documentation/modules/adding-source-provider.adoc
+++ b/documentation/modules/adding-source-provider.adoc
@@ -6,7 +6,7 @@
 ifdef::vmware[]
 = Adding a VMware vSphere source provider
 
-You can add a VMware vSphere source provider by using the {ocp} web console.
+You can migrate VMware vSphere VMs from VMware vCenter or from a VMWare ESX/ESXi server. In {project-short} versions 2.6 and later, you can migrate directly from an ESX/ESXi server, without going through vCenter, by specifying the SDK endpoint to that of an ESX/ESXi server.
 
 [IMPORTANT]
 ====
@@ -15,7 +15,7 @@ EMS enforcement is disabled for migrations with VMware vSphere source providers 
 
 .Prerequisites
 
-* VMware Virtual Disk Development Kit (VDDK) image in a secure registry that is accessible to all clusters.
+* It is strongly recommended to create a VMware Virtual Disk Development Kit (VDDK) image in a secure registry that is accessible to all clusters. A VDDK image accelerates migration. For more information, see xref:../master.adoc#creating-vddk-image_mtv[Creating a VDDK image].
 endif::[]
 ifdef::rhv[]
 = Adding {a-rhv} source provider
@@ -83,10 +83,11 @@ ifdef::vmware[]
 . Specify the following fields:
 
 * *Provider resource name*: Name of the source provider.
+* *Endpoint type*: Select the vSphere provider endpoint type. Options: *vCenter* or *ESXi*. You can migrate virtual machines from vCenter, an ESX/ESXi server that is not managed by vCenter, or from an ESX/ESXi server that is managed by vCenter but does not go through vCenter.
 * *URL*: URL of the SDK endpoint of the vCenter on which the source VM is mounted. Ensure that the URL includes the `sdk` path, usually `/sdk`. For example, `https://vCenter-host-example.com/sdk`. If a certificate for FQDN is specified, the value of this field needs to match the FQDN in the certificate.
 * *VDDK init image*: `VDDKInitImage` path. It is strongly recommended to create a VDDK init image to accelerate migrations. For more information, see xref:../master.adoc#creating-vddk-image_mtv[Creating a VDDK image].
-* *Username*: vCenter user. For example, `user@vsphere.local`.
-* *Password*: vCenter user password.
+* *Username*: vCenter user or ESXi user. For example, `user@vsphere.local`.
+* *Password*: vCenter user password or ESXi user password.
 * *SHA-1 fingerprint*: The provider currently requires the SHA-1 fingerprint of the vCenter Server's TLS certificate in all circumstances. vSphere calls this the server's thumbprint.
 +
 


documentation/modules/adding-source-provider.adoc
```diff
@@ -8,6 +8,8 @@ ifdef::vmware[]
 
 You can add a VMware vSphere source provider by using the {ocp} web console.
 
+In MTV versions 2.6 and later, you can also migrate VMs from standalone ESX/ESXi, or from ESX/ESXi servers that are managed by vCenter without going through vCenter.  This option is included in the procedure that follows.
```
anarnold97, line 11:
```suggestion
In {project-short} versions 2.6 and later, you can also migrate VMs from standalone ESX/ESXi, or from ESX/ESXi servers that are managed by vCenter without going through vCenter.  This option is included in the procedure that follows.
```

documentation/modules/adding-source-provider.adoc
```diff
@@ -78,6 +80,7 @@ ifdef::vmware[]
 . Specify the following fields:
 
 * *Provider resource name*: Name of the source provider.
+* *Endpoint type*: Select the vSphere provider endpoint type. Options: *vSphere* or *ESXi*.
```
yaacov, line 83: it's vCenter or ESXi , @ahadas please keep me honest here

documentation/modules/adding-source-provider.adoc
```diff
@@ -78,6 +80,7 @@ ifdef::vmware[]
...
 * *URL*: URL of the SDK endpoint of the vCenter on which the source VM is mounted. Ensure that the URL includes the `sdk` path, usually `/sdk`. For example, `https://vCenter-host-example.com/sdk`. If a certificate for FQDN is specified, the value of this field needs to match the FQDN in the certificate.
 * *VDDK init image*: `VDDKInitImage` path. It is strongly recommended to create a VDDK init image to accelerate migrations. For more information, see xref:../master.adoc#creating-vddk-image_mtv[Creating a VDDK image].
 * *Username*: vCenter user. For example, `user@vsphere.local`.
```
yaacov, line 86: vCenter and ESXi may have different user name and password, so this should not make users think they need to use there vCenter credentials even if they connect to ESXi, on ESXi they should use the ESXi user+password and on vCenter they should use the vCenter user+password,  , @ahadas please keep me honest here

documentation/modules/adding-source-provider.adoc
RichardHoch: Fixed.

RichardHoch:
@yaacov I made the changes you suggested.
@ahadas Please review this PR as QA.

yaacov:
lgtm

documentation/modules/adding-source-provider.adoc
```diff
@@ -8,6 +8,8 @@ ifdef::vmware[]
 
 You can add a VMware vSphere source provider by using the {ocp} web console.
 
+In {project-short} versions 2.6 and later, you can also migrate VMs from standalone ESX/ESXi, or from ESX/ESXi servers that are managed by vCenter without going through vCenter.  This option is included in the procedure that follows.
```
ahadas, line 11: we don't mention vCenter before so it may not be so obvious to understand what "you can also" refers to
I'm also not sure if for "standalone ESX/ESXi" is a common term
how about:
```suggestion
In {project-short} versions 2.6 and later, in addition to migrations from vCenter, you can also migrate VMs from standalone ESX/ESXi, i.e., an ESX/ESXi server that is not managed by vCenter, or from ESX/ESXi servers that are managed by vCenter without going through vCenter.  This option is included in the procedure that follows.
```

documentation/modules/adding-source-provider.adoc
```diff
@@ -78,10 +80,11 @@ ifdef::vmware[]
 . Specify the following fields:
 
 * *Provider resource name*: Name of the source provider.
+* *Endpoint type*: Select the vSphere provider endpoint type. Options: *vCenter* or *ESXi*.
```
ahadas, line 83: or maybe regarding the previous comment, this is the documentation of 2.6 so we can drop that statement above (or move it to the release notes) and add it here:
 ```suggestion
* *Endpoint type*: Select the vSphere provider endpoint type. Options: *vCenter* or *ESXi*. ESXi can be used for standalone servers, i.e.,  an ESX/ESXi server that is not managed by vCenter, or from ESX/ESXi servers that are managed by vCenter without going through vCenter.
```
what do you think?

File: documentation/modules/adding-source-provider.adoc (hunk 4 above)
Comment (RichardHoch) [reply documentation/modules/adding-source-provider.adoc:11]: If "standalone ESX/ESXi" doesn't sound good to you, then I'll avoid it.  Also the sentence beginning "This option is included ..." is not needed -- since I mentioned the option, I'd better include it in the documentation :) My bad.

File: documentation/modules/adding-source-provider.adoc (hunk 5 above)
Comment (RichardHoch) [reply documentation/modules/adding-source-provider.adoc]: @ahadas , I made some changes based on your suggestions. Please review the PR again.

File: documentation/modules/adding-source-provider.adoc [hunk 6]
```diff
@@ -8,19 +8,21 @@ ifdef::vmware[]
 
 You can add a VMware vSphere source provider by using the {ocp} web console.
 
+In {project-short} versions 2.6 and later, you can migrate VMs from vCenter, from an ESX/ESXi server that is not managed by vCenter, or from an ESX/ESXi server that is managed by vCenter but does not go through vCenter.
```
ahadas, line 11: @RichardHoch that's better but about the last part "but does not go through vCenter" - the idea is that MTV doesn't interact with vCenter in this situation (of a server that is managed by vCenter but the specified endpoint is that of the server). this phrasing makes it sound as the server doesn't go through vCenter

documentation/modules/adding-source-provider.adoc
```diff
@@ -8,19 +8,21 @@ ifdef::vmware[]
...
 .Prerequisites
 
-* VMware Virtual Disk Development Kit (VDDK) image in a secure registry that is accessible to all clusters.
+* It is strongly recommended to create a VMware Virtual Disk Development Kit (VDDK) image in a secure registry that is accessible to all clusters. A VDDK image accelerates migration. For more information, see xref:../master.adoc#creating-vddk-image_mtv[Creating a VDDK image].
 endif::[]
 ifdef::rhv[]
 = Adding {a-rhv} source provider
 
-You can add {a-rhv} source provider by using the {ocp} web console.
+You can add {a-rhv} source provider by using the {ocp } web console.
```
ahadas, line 25:
```suggestion
You can add {a-rhv} source provider by using the {ocp} web console.
```

RichardHoch:
@ahadas I made the chnages you suggested. Please review the PR again.

ahadas (approved):

PR #483, opened 2024-03-04
The Overview page for MTV 2.6

RichardHoch:
MTV 2.6.0

Resolves https://issues.redhat.com/browse/MTV-925 by describing the Overview page for MTV 2.6.0.

Preview: https://file.emea.redhat.com/rhoch/overview_260/html-single/#mtv-overview-page_mtv


RichardHoch:
@yaacov  @ahadas Please review this PR.

documentation/modules/mtv-overview-page.adoc
```diff
@@ -28,10 +55,10 @@ The *Overview* page displays the following information:
 ** Succeeded
 ** Canceled
 
-* Operator: The namespace on which the {project-short} Operator is deployed and the status of the Operator.
+* Chart showing the number of running, failed, and succeeded migrations for each of the last 7 days performed using {project-short}
```
anarnold97, line 58:
```suggestion
* Chart showing the number of running, failed, and succeeded migrations performed using {project-short} for each of the last 7 days.
```

documentation/modules/mtv-overview-page.adoc
```diff
@@ -28,10 +55,10 @@ The *Overview* page displays the following information:
...
+
+* Chart showing the number of running, failed, and succeeded virtual machine migrations for each of the last 7 days performed using {project-short}
```
anarnold97, line 60:
```suggestion
* Chart showing the number of running, failed, and succeeded virtual machine migrations performed using {project-short} for each of the last 7 days. 
```

yaacov:
lgtm, content wise

documentation/modules/mtv-overview-page.adoc
```diff
@@ -10,7 +10,34 @@ The {project-first} *Overview* page displays system-wide information about migra
...
+* Overview
+* YAML
+* Metrics
+
+[id="overview-tab_{context}"]
+== Overview tab
+
+The *Overview* tab lets you see:
+
+* Operator: The namespace on which the {project-short} Operator and the status of the Operator
```
ahadas, line 24:
```suggestion
* Operator: The namespace on which the {project-short} Operator is deployed and the status of the Operator
```
or maybe in which the {project-short} runs?

documentation/modules/mtv-overview-page.adoc
```diff
@@ -10,7 +10,34 @@ The {project-first} *Overview* page displays system-wide information about migra
...
+* Pods: The name, status, and creation time of each pod
```
ahadas, line 25: need to explain what these pods are
```suggestion
* Pods: The name, status, and creation time of each pod that is deployed by the {project-short} Operator
```

documentation/modules/mtv-overview-page.adoc
```diff
@@ -10,7 +10,34 @@ The {project-first} *Overview* page displays system-wide information about migra
...
+* Conditions: Status of the {project-short} Operator:
+
+** Failure: Last failure. `False` indicates no failure since deployment.
+** Running: Whether the Operator is currently running and waiting for the next reconciliation.
+** Successful: Last successful reconciliation.
+
+[id="yaml-tab_{context}"]
+== YAML tab
+
+The last manifest created or retrieved by the CLI is displayed. You can save, reload, or cancel your changes.
```
ahadas, line 35:
```suggestion
The custom resource ForkliftController that defines the operation of the {project-short} Operator. You can modify the custom resource from this tab.
```
it's not very clear what manifest here means so I think it's better to say ForkliftController explicitly
in addition, saying "last" is problematic because it may not show the latest state (and then the user needs to press "reload".

@yaacov it's really nice to have this tab, I keep explaining on calls how to edit this CR from the CLI...

documentation/modules/mtv-overview-page.adoc
```diff
@@ -28,10 +55,10 @@ The *Overview* page displays the following information:
 ** Succeeded
 ** Canceled
 
-* Operator: The namespace on which the {project-short} Operator is deployed and the status of the Operator.
+* Chart showing the number of running, failed, and succeeded migrations performed using {project-short} for each of the last 7 days
```
ahadas, line 58: even to me (and I saw demos of this screen couple of times) the difference between `migrations` here and `virtual machine migrations` below is not clear, does this line refer to migration plans? @yaacov

RichardHoch: @ahadas @yaacov  If a migration migrated 5 machines, wouldn't that add 1 to the migration count and 5 to the VM migration count? Since we have customers migrating tens or hundreds of machines in a single migration, I figured that's what was being counted here. 

I can add an example once we clarify the point.

documentation/modules/mtv-overview-page.adoc
RichardHoch: @ahadas I went with "is deployed" because it might be deployed but not running (right?).

documentation/modules/mtv-overview-page.adoc
RichardHoch: @ahadas I added this note:
[NOTE]
====
Since a single migration might involve many virtual machines, the number of migrations performed by {project-short} might vary significantly from the number of virtual machines that have been migrated by {project-short}.
====

If that's correct, I don't think we need an example, but up to you.

ahadas: @RichardHoch ack, this way (with the note) the term `migration` here is correlated with a migration-plan that can involve multiple VMs and then it makes sense - cool

ahadas (commented):
@RichardHoch thanks
let's enhance the description of the pods that are shown in the Overview tab and this PR will be ready to go from my point of view

RichardHoch:
> @RichardHoch thanks let's enhance the description of the pods that are shown in the Overview tab and this PR will be ready to go from my point of view

@ahadas Is " Pods: The name, status, and creation time of each pod that was deployed by the {project-short} Operator " enough? If you want more info added, please supply it or point me in the right directions. Looks like I forgot to make the change you suggested. My bad. Will fix today.

ahadas:
> > @RichardHoch thanks let's enhance the description of the pods that are shown in the Overview tab and this PR will be ready to go from my point of view
> 
> @ahadas Is " Pods: The name, status, and creation time of each pod that was deployed by the {project-short} Operator " enough? If you want more info added, please supply it or point me in the right directions. 

yeah, that would be enough - just to set the expectation right, that users won't expect to see there all the pods that run in the openshift-mtv namespace

> Looks like I forgot to make the change you suggested. My bad. Will fix today.

right, I noticed that the changes have not been uploaded - I assumed something went wrong with the rebase :)

RichardHoch:
@ahadas Changes made, pushed, and squashed. That should do it. Please review again.

ahadas (approved):

PR #491, opened 2024-03-25
Revised RHV permissions

RichardHoch:
MTV 2.6

Resolves https://issues.redhat.com/browse/MTV-271 by adding additional permissions to the RHV prerequisites.

Preview: https://file.emea.redhat.com/rhoch/rhv_permissions_260/html-single/#rhv-prerequisites_mtv

Note this PR replaces https://github.com/kubev2v/forklift-documentation/pull/426. which Sasha started.

### DIFF
--- a/documentation/modules/rhv-prerequisites.adoc
+++ b/documentation/modules/rhv-prerequisites.adoc
@@ -8,11 +8,23 @@
 
 The following prerequisites apply to {rhv-full} migrations:
 
-* You must use a xref:compatibility-guidelines_{context}[compatible version] of {rhv-full}.
-* You must have the {manager} CA certificate, unless it was replaced by a third-party certificate, in which case, specify the {manager} Apache CA certificate.
+* To create a source provider, you must have at least the `UserRole` and `ReadOnlyAdmin` roles assigned to you. These are the minimum required permissions, however, any other administrator or superuser permissions will also work.
+
+[IMPORTANT]
+====
+You must keep the `UserRole` and `ReadOnlyAdmin` roles until the virtual machines of the source provider have been migrated. Otherwise, the migration will fail.
+====
+
+* To migrate virtual machines:
+** You must have one of the following:
+*** {rhv-short} admin permissions. These permissions allow you to migrate any virtual machine in the system.
+*** `DiskCreator` and `UserVmManager` permissions on every virtual machine you want to migrate.
+
+** You must use a xref:compatibility-guidelines_{context}[compatible version] of {rhv-full}.
+** You must have the {manager} CA certificate, unless it was replaced by a third-party certificate, in which case, specify the {manager} Apache CA certificate.
 +
 You can obtain the {manager}  CA certificate by navigating to https://<engine_host>/ovirt-engine/services/pki-resource?resource=ca-certificate&format=X509-PEM-CA in a browser.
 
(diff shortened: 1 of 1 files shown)


RichardHoch:
@bkhizgiy @ahadas Please review this PR.

documentation/modules/rhv-prerequisites.adoc
```diff
@@ -8,11 +8,18 @@
 
 The following prerequisites apply to {rhv-full} migrations:
 
-* You must use a xref:compatibility-guidelines_{context}[compatible version] of {rhv-full}.
-* You must have the {manager} CA certificate, unless it was replaced by a third-party certificate, in which case, specify the {manager} Apache CA certificate.
+* To create a provider, you must have at least the `UserRole` and `ReadOnlyAdmin` roles assigned to you. These are the minimum required permissions, however, any other administrator or superuser permissions will also work.
+
+* To migrate virtual machines:
```
bkhizgiy, line 13: Looking good overall. The only thing I'm not sure about is if it's clear enough that both sets of permissions had to be given together. The first one is required for the addition of the provider, but it's also important to leave them after for the migration itself.

RichardHoch: @bkhizgiy I don't understand. Do you need to have `UserRole` and `ReadOnlyAdmin` roles as well as RHV admin permissions (and  `DiskCreator` and `UserVmManager` permissions)  to migrate VMs?

bkhizgiy: @RichardHoch I think I didn't explain myself well; sorry for the confusion.

The first set of permissions, `UserRole` and `ReadOnlyAdmin`, are required for the addition of the provider but also need to remain for the migration itself. That means if the user adds these permissions, then adds the provider, and afterward removes the permissions, the migration will fail.

The second set of permissions is required only for the migration and needs to be granted only to the VMs the user wishes to migrate. However, it should be given in addition to the first set of permissions.

I just thought it might not be clear enough from the current phrasing.

RichardHoch:
@bkhizgiy See if this is correct and clear now.
@ahadas Please review as QA.

bkhizgiy:
@RichardHoch Looking good now, I think it's more clear this way, approved by me.

bkhizgiy (approved):

documentation/modules/rhv-prerequisites.adoc
```diff
@@ -8,11 +8,23 @@
...
+
+[IMPORTANT]
+====
+You must keep the `UserRole` and `ReadOnlyAdmin` roles until the virtual machines of the source provider have been migrated. Otherwise, the migration will fail.
+====
+
+* To migrate virtual machines:
+** You must have one of the following:
+*** {rhv-short} admin permissions. These permissions allow you to migrate any virtual machine in the system.
+*** `DiskCreator` and `UserVmManager` permissions. These permissions have to be granted to you individually by an `admin` user per each individual virtual machine that you want to migrate.
```
ahadas, line 21: I think the second part of this sentence is too detailed - it would be ok if the specified user is part of a group that are granted with these permissions (so it's not granted individually for the user) and it's also ok to have permissions on the cluster that propagate to all the vms within the cluster (so it's not individually per VM)

how about replacing this with:
```suggestion
*** `DiskCreator` and `UserVmManager` permissions for the virtual machines you want to update.
```

RichardHoch:
@ahadas I made the change you suggested. please review the PR again.

ahadas (approved):

PR #494, opened 2024-04-13
WIP - MTV-1053: doc fixes for 2.6.0

anarnold97:
### JIRA

* [MTV-1053](https://issues.redhat.com/browse/MTV-1053)


documentation/modules/new-migrating-virtual-machines-cli.adoc
```diff
@@ -42,8 +42,9 @@ endif::[]
 
 .Procedure
 . Create a `Secret` manifest for the source provider credentials:
-+
```
anarnold97, line 45: @RichardHoch - having the indent here means, when the doc build, it *indedents* then breaks out of the indent when it enters the `ifdef` clause. 

So to fix, I propose we move the *indent* inside the `ifdef` clause

documentation/modules/ostack-app-cred-auth.adoc
```diff
@@ -72,6 +72,7 @@ EOF
 
 ** For authentication using the application credential name:
 +
+[source,yaml]
```
anarnold97, line 75: this is just a missing definition for the codeblock... probably made no difference to anyone

documentation/doc-Migration_Toolkit_for_Virtualization/master.adoc
```diff
@@ -180,8 +180,8 @@ include::modules/adding-source-provider.adoc[leveloffset=+4]
 
 include::modules/selecting-migration-network-for-virt-provider.adoc[leveloffset=+4]
 
-include::modules/creating-network-mapping.adoc[leveloffset=+2]
-include::modules/creating-storage-mapping.adoc[leveloffset=+2]
+// include::modules/creating-network-mapping.adoc[leveloffset=+2]
```
anarnold97, line 183: @RichardHoch - i think simply commenting them out is the best solution.

If we had more time, I would use an `ifdef` statement using the `{project-z-version}` variable to reveal at on a future release but this is the quickest solution given the time constraints.

documentation/modules/new-migrating-virtual-machines-cli.adoc
```diff
@@ -306,6 +306,8 @@ spec:
     namespace: <namespace>
 EOF
 ----
+<1> Specify the URL of the API endpoint.
```
anarnold97, line 309: Missing callouts text
image

@RichardHoch - I have the feeling this will fix the duplicate callouts later you note later. What do you think?

documentation/doc-Migration_Toolkit_for_Virtualization/master.adoc
```diff
@@ -237,14 +237,18 @@ include::modules/new-migrating-virtual-machines-cli.adoc[leveloffset=+3]
 :rhv:
 include::modules/new-migrating-virtual-machines-cli.adoc[leveloffset=+3]
 :rhv!:
-:context: ova
-:ova:
-include::modules/new-migrating-virtual-machines-cli.adoc[leveloffset=+3]
-:ova!:
```
anarnold97, line 243: @RichardHoch - Changing the order. Remember that it is not just opening the condition but also closing the condition, otherwise you will get a build error 

image

documentation/modules/ostack-app-cred-auth.adoc
RichardHoch: Good catch! As you say, better safe than sorry!

RichardHoch:
LGTM! Thanks so much!

PR #510, opened 2024-05-02
Attributes for MTV 2.6.1

RichardHoch:
MTV 2.6.1

Resolves https://issues.redhat.com/browse/MTV-1094 by changing document attributes where needed.

### DIFF
--- a/documentation/modules/common-attributes.adoc
+++ b/documentation/modules/common-attributes.adoc
@@ -8,7 +8,7 @@
 :ocp: Red Hat OpenShift
 :ocp-name: OpenShift
 :ocp-short: OpenShift
-:ocp-version: 4.11
+:ocp-version: 4.14
 :operator: mtv-operator
 :operator-name-ui: Migration Toolkit for Virtualization Operator
 :operator-name: MTV Operator
@@ -17,7 +17,7 @@
 :project-short: MTV
 :project-first: {project-full} ({project-short})
 :project-version: 2.6
-:project-z-version: 2.6.0
+:project-z-version: 2.6.1
 :the: The
 :the-lc: the
 :virt: OpenShift Virtualization
--- a/documentation/modules/creating-vddk-image.adoc
+++ b/documentation/modules/creating-vddk-image.adoc
@@ -35,11 +35,14 @@ $ mkdir /tmp/<dir_name> && cd /tmp/<dir_name>
 
 . In a browser, navigate to the link:https://developer.vmware.com/web/sdk/8.0/vddk[VMware VDDK version 8 download page].
 . Select version 8.0.1 and click *Download*.
-+
+
 [NOTE]
 ====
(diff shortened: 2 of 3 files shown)


RichardHoch:
@ahadas Please review.

documentation/modules/common-attributes.adoc
```diff
@@ -8,7 +8,7 @@
 :ocp: Red Hat OpenShift
 :ocp-name: OpenShift
 :ocp-short: OpenShift
-:ocp-version: 4.11
+:ocp-version: 4.15
```
ahadas, line 11: I see that `ocp-version` is used mainly in links to the OpenShift documentation but also in the prerequisites of the MTV operator: "* {ocp} {ocp-version} or later installed." - and in that case, it should be the earliest supported version, which in case of 2.6 is 4.14

documentation/modules/creating-vddk-image.adoc
```diff
@@ -35,11 +35,6 @@ $ mkdir /tmp/<dir_name> && cd /tmp/<dir_name>
 
 . In a browser, navigate to the link:https://developer.vmware.com/web/sdk/8.0/vddk[VMware VDDK version 8 download page].
 . Select version 8.0.1 and click *Download*.
-+
-[NOTE]
-====
-In order to migrate to {virt} 4.12, download VDDK version 7.0.3.2 from the link:https://developer.vmware.com/web/sdk/7.0/vddk[VMware VDDK version 7 download page].
-====
```
ahadas, line 42: as I commented on slack, we better keep this note as long as 4.12 is supported - how about adding a comment to remove it once 4.12 becomes unsupported instead?

documentation/modules/installing-mtv-operator.adoc
```diff
@@ -17,7 +17,7 @@ endif::[]
 
 .Prerequisites
 
-* {ocp} {ocp-version} or later installed.
+* {ocp} 4.14 or 4.15 later installed.
```
ahadas, line 20: ah ok, I referred to this part in my previous comment about `ocp-version` - I think it's better to keep referencing `ocp-version` and set it to the earliest supported version, @RichardHoch what do you think?

documentation/modules/common-attributes.adoc
RichardHoch: Good thinking. Fixed.

documentation/modules/installing-mtv-operator.adoc
RichardHoch: @ahadas Agreed .

documentation/modules/creating-vddk-image.adoc
RichardHoch: Done.

RichardHoch:
@ahadas I made the changes you suggested. Please review the PR again.

ahadas (approved):

PR #532, opened 2024-07-17
NFC-enabled vmkernel NIC migration option

RichardHoch:
MTV 2.6.3

Resolves https://issues.redhat.com/browse/MTV-1200 by adding a note to the MTV UI and CLI migration procedures about configuring a host on the vmWare site, rather than via MTV.  

Previews:

- https://file.emea.redhat.com/rhoch/vmware_esxi_nfc/html-single/#adding-source-providers [see note in section 4.4.1.1.1, "Selecting a migration network for a VMware source provider"
- https://file.emea.redhat.com/rhoch/vmware_esxi_nfc/html-single/#new-migrating-virtual-machines-cli_vmware [see note at the end of step 3]

### DIFF
--- a/documentation/modules/snip_vmware_esxi_nfc.adoc
+++ b/documentation/modules/snip_vmware_esxi_nfc.adoc
@@ -1,0 +1,6 @@
+:_content-type: SNIPPET
+
+[NOTE]
+====
+You can also control the network from which disks are transferred from a host by using the Network File Copy (NFC) service in vSphere.
+====
--- a/documentation/modules/new-migrating-virtual-machines-cli.adoc
+++ b/documentation/modules/new-migrating-virtual-machines-cli.adoc
@@ -354,7 +354,9 @@ EOF
 ----
 <1> Specify the name of the VMware vSphere `Provider` CR.
 <2> Specify the Managed Object Reference (MoRef) of the VMware vSphere host.
-<3> Specify the IP address of the VMware vSphere migration network.
+<3> Specify the IP address of the VMware vSphere migration network
+
+include::snip_vmware_esxi_nfc.adoc[]
 
 [start=4]
 . Create a `NetworkMap` manifest to map the source and destination networks:
--- a/documentation/modules/selecting-migration-network-for-vmware-source-provider.adoc
+++ b/documentation/modules/selecting-migration-network-for-vmware-source-provider.adoc
@@ -10,6 +10,8 @@ You can select a migration network in the {ocp} web console for a source provide
 
 Using the default network for migration can result in poor performance because the network might not have sufficient bandwidth. This situation can have a negative effect on the source platform because the disk transfer operation might saturate the network.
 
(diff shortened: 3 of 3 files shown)


anarnold97 (approved):
LGTM

RichardHoch:
@ahadas Please review.

documentation/modules/snip_vmware_esxi_nfc.adoc
```diff
@@ -0,0 +1,6 @@
+:_content-type: SNIPPET
+
+[NOTE]
+====
+If you are using an EXSi host, you can configure the `vSphere Backup NFC` on the VMware site instead of specifying a host in your {project-short} migration plan.
```
ahadas, line 5:
```suggestion
If you are using an ESXi host, you can configure `vSphere Backup NFC` on the VMware side instead of specifying a host in your {project-short} migration plan.
```
the changes above address phrasing/typo thingies but I also think we can drop the `If you are using an ESXi host` since ESXi hosts are always involved in migrations from vSphere? (or @RichardHoch , maybe I misunderstood what's the purpose of this part)
and also `instead of specifying a host in our MTV migration plan` - the transfer network on the VMware side can only be set at the host level, not at the plan level, no? (in that case, we need to drop it from documentation/modules/creating-migration-plan-2-6-3.adoc)

RichardHoch:
@ahadas  David V. approved the changed text. Please give it a look (as QA).
@anarnold97 Please give the new text a look.
Please and thanks.

documentation/modules/creating-migration-plan-2-6-3.adoc
```diff
@@ -55,6 +55,8 @@ To specify a different root device, in the *Settings* section, click the Edit ic
 {project-short} uses the following format for disk location: `/dev/sd<disk_identifier><disk_partition>`. For example, if the second disk is the root device and the operating system is on the disk's second partition, the format would be: `/dev/sdb2`. After you enter the boot device, click *Save*.
 +
 If the conversion fails because the boot device provided is incorrect, it is possible to get the correct information by looking at the conversion pod logs.
++
+include::snip_vmware_esxi_nfc.adoc[]
```
ahadas, line 59: I think this one should drop since we don't mention setting the ESXi network here (right?)

RichardHoch: Removed.

RichardHoch: Removed.

RichardHoch:
@ahadas I made the changes you suggested. Please look at the previews to be sure you like what you see. For now, I have included previews of the procedures for migration via the UI [2nd and 3rd bullets] so you can see what they look like without the note. If you still want the note out, I'll remove those 2 previews.

ahadas (approved):

anarnold97 (approved):
LGTM

PR #540, opened 2024-08-11
WIP MTV-1316: Removing VMware tools from Windows guests

anarnold97:


anarnold97:
@ahadas - please can I ask if you are ok with this, before i get a second review from Richard Jones?

Thanks

PR #552, opened 2024-09-10
preserve static ip is for all families, not only windows

fabiand:


documentation/modules/creating-migration-plan-2-6-3.adoc
```diff
@@ -42,7 +42,7 @@ The *Create migration plan* pane opens. It displays the source provider's name a
 
 . VMware source providers only (All optional):
 
-* *Preserving static IPs of Windows VMs*: By default, virtual network interface controllers (vNICs) change during the migration process. This results in vNICs that are set with a static IP in vSphere losing their IP. To avoid this by preserving static IPs of Windows VMs, click the Edit icon next to *Preserve static IPs*, and toggle the *Whether to preserve the static IPs* switch in the window that opens. Then click *Save*.
+* *Preserving static IPs of VMs*: By default, virtual network interface controllers (vNICs) change during the migration process. This results in vNICs that are set with a static IP in vSphere losing their IP. To avoid this by preserving static IPs of VMs, click the Edit icon next to *Preserve static IPs*, and toggle the *Whether to preserve the static IPs* switch in the window that opens. Then click *Save*.
```
anarnold97, line 45:
```suggestion
* *Preserving static IPs of VMs*: By default, virtual network interface controllers (vNICs) change during the migration process. This results in vNICs that are set with a static IP in vSphere losing their IP. To avoid this by preserving static IPs of VMs, click the Edit icon next to *Preserve static IPs* and toggle the *Whether to preserve the static IPs* switch in the window that opens. Then click *Save*.
```

PR #559, opened 2024-09-22
Additions to cold vs warm migration

RichardHoch:
MTV 2.7.0

Resolves https://issues.redhat.com/browse/MTV-1254 by adding more information about the differences between cold and warm migrations for MTV 2.7.0.

Preview: https://file.corp.redhat.com/rhoch/warm_col/html-single/#warm-migration-versus-cold-migration [section 1.1.3]

### DIFF
--- a/documentation/modules/about-cold-warm-migration.adoc
+++ b/documentation/modules/about-cold-warm-migration.adoc
@@ -31,7 +31,7 @@ Then the VMs are shut down and the remaining data is copied during the _cutover_
 
 The VMs are not shut down during the precopy stage.
 
-The VM disks are copied incrementally using link:https://kb.vmware.com/s/article/1020128[changed block tracking (CBT)] snapshots. The snapshots are created at one-hour intervals by default. You can change the snapshot interval by updating the `forklift-controller` deployment.
+The VM disks are copied incrementally by using link:https://kb.vmware.com/s/article/1020128[changed block tracking (CBT)] snapshots. The snapshots are created at one-hour intervals by default. You can change the snapshot interval by updating the `forklift-controller` deployment.
 
 [IMPORTANT]
 ====
@@ -47,3 +47,72 @@ The precopy stage runs until the cutover stage is started manually or is schedul
 The VMs are shut down during the cutover stage and the remaining data is migrated. Data stored in RAM is not migrated.
 
 You can start the cutover stage manually by using the {project-short} console or you can schedule a cutover time in the `Migration` manifest.
+
+
+[id="warm-migration-versus-cold-migration"_{context}]
+== Advantages and disadvantages of cold and warm migrations
+
+[id="warm-cold-migration-overview_{context}"]
+=== Overview
+
+include::snip_cold-warm-comparison-table.adoc[]
+
+[id="warm-cold-migration-detailed_{context}"]
+=== Detailed description
+
+The table that follows offers a more detailed description of the advantages and disadvantages of each type of migration. It assumes that you have installed Red Hat Enterprise Linux (RHEL) 9 on the {ocp} platform on which you installed {project-short}.
+
+[cols="1,1,1",options="header"]
+.Detailed description of advantages and disadvantages
+
+|===
+| |Cold migration |Warm migration
+
+|*Fail fast*
+|Each VM is converted to be compatible with {ocp-short} and, if the conversion is successful, the VM is transferred. If a VM cannot be converted, the migration fails immediately.
+|For each VM, {project-short} creates a snapshot and transfers it to {ocp}. When you start the cutover, {project-short} creates the last snapshot, transfers it, and then converts the VM.
+
+|*Tools*
+|{project-short} only.
+|{project-short} and CDI from {virt}.
+
+|*Parallelism*
+|Disks must be transferred sequentially.
+|Disks can be transferred in parallel, using different pods.
+|===
+
+[NOTE]
+====
+The preceding table describes the situation for VMs that are running, because the main benefit of warm migration is the reduced downtime, and there is no reason to initiate warm migration for VMs that are down. However, performing warm migration for VMs that are down is not the same as cold migration, even when {project-short} uses `virt-v2v` and RHEL 9. For VMs that are down, {project-short} transfers the disks using CDI, unlike in cold migration.
+====
+
+[NOTE]
+====
+When importing from VMware, there are additional factors which impact the migration speed such as limits related to ESXi, vSphere. or VDDK.
+====
+
+=== Conclusions
+
+Based on the preceding information, we can draw the following conclusions about cold migration vs. warm migration:
+
+* The shortest downtime of VMs can be achieved by using warm migration.
+* The shortest duration for VMs with a large amount of data on a single disk can be achieved by using cold migration.
+* The shortest duration for VMs with a large amount of data that is spread evenly across multiple disks can be achieved by using warm migration.
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
--- a/documentation/modules/about-cold-warm-migration.adoc
+++ b/documentation/modules/about-cold-warm-migration.adoc
@@ -78,7 +78,7 @@ The table that follows offers a more detailed description of the advantages and 
 
 |*Parallelism*
 |Disks must be transferred sequentially.
-|Disks can be transferred in parallel, using different pods.
+|Disks can be transferred in parallel using different pods.
 |===
 
 [NOTE]
--- a/documentation/modules/about-cold-warm-migration.adoc
+++ b/documentation/modules/about-cold-warm-migration.adoc
@@ -83,7 +83,7 @@ The table that follows offers a more detailed description of the advantages and 
 
 [NOTE]
 ====
-The preceding table describes the situation for VMs that are running, because the main benefit of warm migration is the reduced downtime, and there is no reason to initiate warm migration for VMs that are down. However, performing warm migration for VMs that are down is not the same as cold migration, even when {project-short} uses `virt-v2v` and RHEL 9. For VMs that are down, {project-short} transfers the disks using CDI, unlike in cold migration.
+The preceding table describes the situation for VMs that are running because the main benefit of warm migration is the reduced downtime, and there is no reason to initiate warm migration for VMs that are down. However, performing warm migration for VMs that are down is not the same as cold migration, even when {project-short} uses `virt-v2v` and RHEL 9. For VMs that are down, {project-short} transfers the disks using CDI, unlike in cold migration.
 ====
 
 [NOTE]
--- a/documentation/modules/snip_cold-warm-comparison-table.adoc
+++ b/documentation/modules/snip_cold-warm-comparison-table.adoc
@@ -1,0 +1,22 @@
+
+:_content-type: SNIPPET
+
+Both cold migration and warm migration have advantages and disadvantages, as described in the table that follows:
+
+[cols="1,1,1",options="header"]
+.Advantages and disadvantages of cold and warm migrations
+|===
+| |Cold migration |Warm migration
+
+|*Duration*
+|Correlates to the amount of data on the disks
+|Correlates to the amount of data on the disks and VM utilization
+
+|*Data transferred*
+|Approximate sum of all disks
+|Approximate sum of all disks and VM utilization
+
+|*VM downtime*
+|High
+|Low
+|===


nunzy1 (commented):
Hi Richard,
A couple of very minor comments. 

Typo: "and hen shut down".

This title caught my eye:  "Cold migration vs. warm migration: Which is right for your migration?"  If this is good with our standards and you want to keep it, all good.  Maybe something like, "Advantages and disadvantages of cold and warm migration"?

RichardHoch:
@nunzy1 Fixed the typo, and went with your suggested title for the section.

documentation/modules/about-cold-warm-migration.adoc
```diff
@@ -47,3 +47,51 @@ The precopy stage runs until the cutover stage is started manually or is schedul
 The VMs are shut down during the cutover stage and the remaining data is migrated. Data stored in RAM is not migrated.
 
 You can start the cutover stage manually by using the {project-short} console or you can schedule a cutover time in the `Migration` manifest.
+
+[id="warm-migration-versus-cold-migration"_{context}]
+== Cold migration vs. warm migration: Which is right for your migration?
+
+Both warm migration and cold migration have advantages and limitations. In general, warm migrations can take longer than cold migrations, but warm migrations involve less downtime than cold migrations.
```
fabiand, line 54: Can we convert the data of this sentence into a table like the following


|  | Duration | VM Downtime |
|---|---|---|
| Cold Migration | Minimal | High |
| Warm migration | Longer and variable | Minimal |

documentation/modules/about-cold-warm-migration.adoc
```diff
@@ -47,3 +47,51 @@ The precopy stage runs until the cutover stage is started manually or is schedul
...
+|===
+| |Cold migration |Warm migration
+
+|*Duration*
+| Each block is copied only once, making the duration of the migration shorter.
+| A block might be copied in more than one snapshot. {project-short} ensures that only the most recent snapshot is used, but eliminating duplicate snapshots adds time to the duration of the migration.
+
+|*Fail fast*
+| Each block is converted to be compatible with {ocp-short} and, if the conversion is successful, the block is transferred. If a block cannot be converted, the migration fails immediately.
+| All blocks are transferred before any are converted. This means that it takes longer for a warm migration to fail than a cold migration, because a cold migration fails before all blocks are transferred.
```
fabiand, line 70: @mnecas thoughts on this?

documentation/modules/about-cold-warm-migration.adoc
```diff
@@ -47,3 +47,51 @@ The precopy stage runs until the cutover stage is started manually or is schedul
...
+
+|*Tools*
+| RHEL 9 and `virt-v2v` in all cases.
```
fabiand, line 73: RHEL 9 and virt-v2v are internal details, thus what about simply saying "MTV only"

documentation/modules/about-cold-warm-migration.adoc
```diff
@@ -47,3 +47,51 @@ The precopy stage runs until the cutover stage is started manually or is schedul
...
+| RHEL 9 and `virt-v2v` for all migrations except those that use the Containerized Data Importer (CDI).
```
fabiand, line 74: Here: MTV and CDI from OpenShift Virtualization

documentation/modules/about-cold-warm-migration.adoc
```diff
@@ -47,3 +47,51 @@ The precopy stage runs until the cutover stage is started manually or is schedul
...
+
+|*VM downtime*
+| Longer than with a warm migration because in a cold migration, all the source VMs are shut down and then the data is transferred.
+| Shorter than with a cold migration, because the disks are transferred, and hen shut down before being converted.
+
+|*Parallelism*
+| Disks must be transferred sequentially.
```
fabiand, line 81: @mnecas really?

documentation/modules/about-cold-warm-migration.adoc
mnecas:
```
Each block is converted to be compatible with {ocp-short} and, if the conversion is successful, the block is transferred. If a block cannot be converted, the migration fails immediately.
```
We do not convert each block. The warm migration in the MTV perspective means creating a snapshot and transferring the snapshot. Once there is a cutover we create the last snapshot, which consists only of the changes between the previous snapshot and the last one. For that we use and **require** Changed Block Tracking (CBT) on the VM and VM disk to be enabled. Once the last snapshot is transferred we do commit the layers and run the guest conversion on the resulting image. After the guest conversion is finished we do create the VM and start it.
Something like this:
image

documentation/modules/about-cold-warm-migration.adoc
mnecas: no,
in warm migration for each disk we create separate DataVolume which imports the `vmdk`, so for each disk we have separate CDI import

RichardHoch: @mnecas "Disks can be transferred in parallel, using different pods" is for warm migration -- I will change this to fit your description.

mnecas: ah the cold migration, yeah virt-v2v transfers them sequentially.

mnecas: @fabiand another reason to synchronize the flows.

documentation/modules/about-cold-warm-migration.adoc
mnecas: I would not say each block. It fails faster because it could fail before the disk transfer. Please take a look at:
https://access.redhat.com/articles/1353223

Unlike warm migrations we are using the virt-v2v directly for both disk transfer and guest conversion

RichardHoch: @anarnold97 As you can see above, I'm going to need to mention the requirement for CBT.  Martin requested I link to this article https://knowledge.broadcom.com/external/article/320557/changed-block-tracking-cbt-on-virtual-ma.html , but I see that it includes a procedure. What do you think about a sentence like "Warm migration requires Changed Block Tracking (CBT), which is described in VMware documentation" and leave it at that, even though that  forces the user to do the research?

mnecas: We don't need the article, we need to say that we **require** the feature and say how to enable it on the VM, which is described in the article solution. ie add advanced parameters:
```
ctkEnabled = TRUE
scsi0:0.ctkEnabled = TRUE
....
scsiX:X.ctkEnabled = TRUE
```

RichardHoch:
@mnecas I just noticed that in the existing documentation, we already mention CBT and even link to the article you suggested: https://docs.redhat.com/en/documentation/migration_toolkit_for_virtualization/2.6/html-single/installing_and_using_the_migration_toolkit_for_virtualization/index#warm-migration_mtv  [see the pre-copy stage]. 

Good enough?

mnecas:
yeah sounds good to me

RichardHoch:
@mnecas I made the changes we discussed. please review this PR again. Thanks.

documentation/modules/about-cold-warm-migration.adoc
```diff
@@ -47,3 +47,83 @@ The precopy stage runs until the cutover stage is started manually or is schedul
...
+
+|Cold migration
+|Minimal
+|High
+
+|Warm migration
+|Longer and variable
+|Minimal
+|===
+
```
anarnold97, line 74: @RichardHoch is there any way you could convert this to a snippet? WDYT

i might want to use it in the release notes

documentation/modules/about-cold-warm-migration.adoc
```diff
@@ -47,3 +47,83 @@ The precopy stage runs until the cutover stage is started manually or is schedul
...
+[cols="1,1,1",options="header"]
+.Detailed description of advantages and disadvantages
+
+|===
+| |Cold migration |Warm migration
+
+|*Duration*
+| Each VM is copied only once, making the duration of the migration shorter.
+| A VM might be copied in more than one snapshot. {project-short} ensures that only the most recent snapshot is used, but eliminating duplicate snapshots adds time to the duration of the migration.
```
mnecas, line 88: This might imply that we are copying the whole disk multiple times but we transfer the disk only once and then we transfer the changed blocks which are made on top of the previous snapshot. This is the main advantage of the warm migration. We don't need to transfer everything after VM is shut down. 

How about:
```
For each VM, MTV creates a snapshot which is transferred. After the cutover, MTV creates a final snapshot which contains only changed blocks between the previous snapshot. This reduces the downtime because the MTV transfers only changes not the whole disks. 
```
wdyt @RichardHoch @fabiand ?

documentation/modules/about-cold-warm-migration.adoc
fabiand: For this place I'd keep a table, as it's easier to parse IMO

documentation/modules/about-cold-warm-migration.adoc
fabiand: Yes, I learned something new with this table.

I think this comment can now be closed.

fabiand (commented):
lgtm!

fabiand (approved):

documentation/modules/about-cold-warm-migration.adoc
RichardHoch: @fabiand He's asking me to create a document snippet containing the table, which he will then use in the release notes -- if I can do that, it will look the same to readers. if I can't, I'll leave the file as is. No worries!

documentation/modules/about-cold-warm-migration.adoc
```diff
@@ -47,3 +47,83 @@ The precopy stage runs until the cutover stage is started manually or is schedul
...
+The table that follows offers a more detailed description of the advantages and disadvantages of each type of migration. It assumes that you have installed Red Hat Enterprise Linux (RHEL) 9 on the {ocp} platform on which you installed {project-short}.
+
+[cols="1,1,1",options="header"]
+.Detailed description of advantages and disadvantages
+
+|===
+| |Cold migration |Warm migration
+
+|*Duration*
+| Each VM is copied only once, making the duration of the migration shorter.
```
fabiand, line 87: Relates to the comment below

> Each disk of a VM is copied only once, leading to the shortest duration for the migration.

documentation/modules/about-cold-warm-migration.adoc
RichardHoch: @mnecas Fine with me. @fabiand WDYT?

fabiand: Can we maybe simplify it:

> Each disk of a VM is first copied once, and then partially transferred again to compensate for change blocks while the VM keeps running leading to a longer and varying duration for the migration.

documentation/modules/about-cold-warm-migration.adoc
```diff
@@ -47,3 +47,83 @@ The precopy stage runs until the cutover stage is started manually or is schedul
...
+[id="warm-cold-migration-overview_{context}"]
+=== Overview
+
+Both cold migration and warm migration have advantages and disadvantages, as described in the table that follows:
+
+[cols="1,1,1",options="header"]
+.Advantages and disadvantages of cold and warm migrations
+
+|===
+| |Duration |VM downtime
```
fabiand, line 64: We need to update this table

Method | Duration | Data transferred | VM Downtime
--|--|--|--
Cold migration | Correlates to amount of disk data | Approx sum of all disks | High
Warm migration | Correlates to amount of disk data and VM utilization | Approx sum of all disks and changed blocks | Low

documentation/modules/about-cold-warm-migration.adoc
```diff
@@ -47,3 +47,83 @@ The precopy stage runs until the cutover stage is started manually or is schedul
...
+
+The table that follows offers a more detailed description of the advantages and disadvantages of each type of migration. It assumes that you have installed Red Hat Enterprise Linux (RHEL) 9 on the {ocp} platform on which you installed {project-short}.
+
+[cols="1,1,1",options="header"]
+.Detailed description of advantages and disadvantages
+
+|===
+| |Cold migration |Warm migration
+
+|*Duration*
```
fabiand, line 86: Cold
Duration: Driven by amount of data on disks
Data transferred: Approx sum of all disks
…

Warm
Duration: Driven by amount of data on disks and VM utilization
Data transferred: Approx sum of all disks and VM utilization
…

documentation/modules/about-cold-warm-migration.adoc
```diff
@@ -47,3 +47,83 @@ The precopy stage runs until the cutover stage is started manually or is schedul
...
+
+|*VM downtime*
+| Longer than with a warm migration because in a cold migration, all the source VMs are shut down and then the data is transferred.
+| Shorter than with a cold migration, because the disks are transferred, and then shut down before being converted.
+
+|*Parallelism*
+| Disks must be transferred sequentially.
+| Disks can be transferred in parallel, using different pods.
+|===
+
```
fabiand, line 106: Can we add one more section

> Conclusion
> Lowest downtime of VMs can be achieved with warm migration.
> Lowest duration for VMs with a lot of data on  a single disk can be achieved with cold migration.
> Lowest duration of VMs with a lot of data spread evenly accross multiple disks can be achieved with warm migration.

fabiand (commented):
We have to rework the duration.
We have to rework it, because duration depends on data transferred and parallelism.
This is way we neeed to move awayy from "short" and "long".
Instead we can say that the duration is "static"/predictable or "dynamic"/unpredictable.
We can then speak about how much data is getting transferred.

documentation/modules/about-cold-warm-migration.adoc
```diff
@@ -47,3 +47,83 @@ The precopy stage runs until the cutover stage is started manually or is schedul
...
+[NOTE]
+====
+The preceding table describes the situation for VMs that are running, because the main benefit of warm migration is the reduced downtime, and there is no reason to initiate warm migration for VMs that are down. However, performing warm migration for VMs that are down is not the same as cold migration, even when {project-short} uses `virt-v2v` and RHEL 9. For VMs that are down, {project-short} transfers the disks using CDI, unlike in cold migration.
+====
+
```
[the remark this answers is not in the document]
RichardHoch: OK

documentation/modules/about-cold-warm-migration.adoc
RichardHoch: OK

documentation/modules/about-cold-warm-migration.adoc
RichardHoch: OK

documentation/modules/about-cold-warm-migration.adoc
RichardHoch: @fabiand  I added the Data transferred rows and changed the contents of the rows as you suggested; I did not remove any rows.

RichardHoch:
@mnecas @fabiand I made the changes you suggested. Please review the PR again.

documentation/modules/about-cold-warm-migration.adoc
RichardHoch: @mnecas Please check now. I forgot to push my changes.

documentation/modules/about-cold-warm-migration.adoc
```diff
@@ -47,3 +47,99 @@ The precopy stage runs until the cutover stage is started manually or is schedul
...
+Both cold migration and warm migration have advantages and disadvantages, as described in the table that follows:
+
+[cols="1,1,1",options="header"]
+.Advantages and disadvantages of cold and warm migrations
+
+|===
+| |Duration |VM downtime
+
+|Cold migration
```
fabiand, line 66: This table was not updated, but should.

RichardHoch:
@fabiand @mnecas I've made the changes you requested. There's some overlap/conflict between the first and second tables -- please check them.

RichardHoch:
@mnecas Please review the PR. Should we remove rows from the detailed chart that are in the small chart?

mnecas (commented):
I'm bit lost about the tables as I see 2 tables with some identical information

RichardHoch:
@mnecas We wanted to have 2 tables -- one more technical than the other. Do you think we should have just one or should i remove the repeats from the second table?

mnecas:
Ah if it was discussed before ok, from technical perspective lgtm

RichardHoch:
@mnecas Made the changes we discussed, rotated the first table so both have "Cold migration" and "Warm migration" as the column names. Please review.

mnecas (approved):

fabiand:
lgtm!

Please merge

fabiand (approved):

documentation/modules/about-cold-warm-migration.adoc
```diff
@@ -47,3 +47,72 @@ The precopy stage runs until the cutover stage is started manually or is schedul
...
+|Each VM is converted to be compatible with {ocp-short} and, if the conversion is successful, the VM is transferred. If a VM cannot be converted, the migration fails immediately.
+|For each VM, {project-short} creates a snapshot and transfers it to {ocp}. When you start the cutover, {project-short} creates the last snapshot, transfers it, and then converts the VM.
+
+|*Tools*
+|{project-short} only.
+|{project-short} and CDI from {virt}.
+
+|*Parallelism*
+|Disks must be transferred sequentially.
+|Disks can be transferred in parallel, using different pods.
```
anarnold97, line 81:
```suggestion
|Disks can be transferred in parallel using different pods.
```

documentation/modules/about-cold-warm-migration.adoc
```diff
@@ -47,3 +47,72 @@ The precopy stage runs until the cutover stage is started manually or is schedul
...
+|===
+
+[NOTE]
+====
+The preceding table describes the situation for VMs that are running, because the main benefit of warm migration is the reduced downtime, and there is no reason to initiate warm migration for VMs that are down. However, performing warm migration for VMs that are down is not the same as cold migration, even when {project-short} uses `virt-v2v` and RHEL 9. For VMs that are down, {project-short} transfers the disks using CDI, unlike in cold migration.
```
anarnold97, line 86:
```suggestion
The preceding table describes the situation for VMs that are running because the main benefit of warm migration is the reduced downtime, and there is no reason to initiate warm migration for VMs that are down. However, performing warm migration for VMs that are down is not the same as cold migration, even when {project-short} uses `virt-v2v` and RHEL 9. For VMs that are down, {project-short} transfers the disks using CDI, unlike in cold migration.
```

PR #563, opened 2024-10-02
MTV-1308: VDDK is mentioned as strongly recommended but in many situations, when not installed, migration fails

anarnold97:
### JIRA

* [MTV-1308](https://issues.redhat.com/browse/MTV-1308)

### PREVIEW

* [4.4.1.1. Adding a VMware vSphere source provider](https://anarnold97.github.io/MTA-Preview/MTV/MTV-1308.html#adding-source-providers)


fabiand (approved):
Looks no less than awesome and good to go.

PR #616, opened 2025-01-14
MTV-1884: Software Compatibility Guideline Inconsistency

anarnold97:
### JIRA

* [MTV-1884](https://issues.redhat.com/browse/MTV-1884)

### PREVIEW

* [Software compatibility guidelines](https://anarnold97.github.io/MTA-Preview/MTV/mtv-version.html#compatibility-guidelines_mtv)

image

### DIFF
--- a/documentation/modules/compatibility-guidelines.adoc
+++ b/documentation/modules/compatibility-guidelines.adoc
@@ -8,6 +8,8 @@
 
 You must install compatible software versions.
 
+
+// check OCP version
 [cols="1,1,1,1,1,1", options="header"]
 .Compatible software versions
 |===
--- a/documentation/modules/compatibility-guidelines.adoc
+++ b/documentation/modules/compatibility-guidelines.adoc
@@ -14,7 +14,7 @@ You must install compatible software versions.
 .Compatible software versions
 |===
 |{project-full} |{ocp} |{virt} |VMware vSphere |{rhv-full}|OpenStack
-|{project-z-version} |4.15, 4.16, and 4.17 |{ocp-version} or later |6.5 or later |4.4 SP1 or later|16.1 or later
+|{project-z-version} |4.15, 4.16, and 4.17 |4.15, 4.16, and 4.17 |6.5 or later |4.4 SP1 or later|16.1 or later
 |===
 
 [NOTE]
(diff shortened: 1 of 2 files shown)


documentation/modules/compatibility-guidelines.adoc
```diff
@@ -8,11 +8,13 @@
 
 You must install compatible software versions.
 
+
+// check OCP version
 [cols="1,1,1,1,1,1", options="header"]
 .Compatible software versions
 |===
 |{project-full} |{ocp} |{virt} |VMware vSphere |{rhv-full}|OpenStack
-|{project-z-version} |{ocp-version} or later |{ocp-version} or later |6.5 or later |4.4 SP1 or later|16.1 or later
```
mnecas, line 15: I'm not sure if we should have this in docs as we will need to maintain it for longer time/additional versions.
If we go with this, we should use the Y stream version instead as the whole Y determines the OCP support? or is this whole page Z stream based?

anarnold97: @mnecas - good point. but I could create an attribute, and then it is updated on Y streams when we update the attributes:

image

anarnold97: @RichardHoch - WDYT. This is pretty maintainable, right?

RichardHoch: @anarnold97 Definitely maintainable.

mnecas: I ment to replace the `project-z-version` with something like `project-y-version` as the ocp versions apply for all Z versions

documentation/modules/compatibility-guidelines.adoc
```diff
@@ -8,11 +8,13 @@
...
+|{project-z-version} |{ocp-y-version} |{ocp-y-version} |6.5 or later |4.4 SP1 or later|16.1 or later
```
anarnold97, line 17:
```suggestion
|{project-version} |{ocp-y-version} |{ocp-y-version} |6.5 or later |4.4 SP1 or later|16.1 or later
```

documentation/modules/compatibility-guidelines.adoc
anarnold97: Oh that is even easier. Swap to this attribute: 

```
:project-version: 2.7
```

anarnold97: image

mnecas: yeah, this looks good to me

mnecas (approved):

PR #578, opened 2024-10-28
Supported migration network for MTV

RichardHoch:
MTV 2.8.0

Resolves https://issues.redhat.com/browse/MTV-607 by adding notes and additions to procedures about supported migration networks for MTV.  

Previews:

- https://file.corp.redhat.com/rhoch/supp_mig_network/html-single/#adding-source-providers [Second note in "4.4.1.1.1. Selecting a migration network for a VMware source provider"]
- https://file.corp.redhat.com/rhoch/supp_mig_network/html-single/#adding-destination-providers [Second note in "4.4.2.2. Selecting a migration network for an OpenShift Virtualization provider"]
- https://file.corp.redhat.com/rhoch/supp_mig_network/html-single/#creating-migration-plan-2-7_provider [Step 8, last bullet]
- https://file.corp.redhat.com/rhoch/supp_mig_network/html-single/#creating-migration-plan-2-7_vms [Step 8, last bullet]

PR #639, opened 2025-02-24
Snippet fix

anarnold97:
@RichardHoch  - here is the snippet fix.

Basically, you would need to do create an xref with the `{context}` already defined

```
ifdef::vmware,dest_vwmware[]

[NOTE]
====
If you input any value of maximum transmission unit (MTU) besides the default value in your migration network, you must also input the same value in the {ocp-short} transfer network that you use. For more information about the {ocp-short} transfer network, see xref:creating-migration-plan-2-8_vmware[Creating a migration plan].
====

endif::[]

ifdef::dest_rhv[]

[NOTE]
====
If you input any value of maximum transmission unit (MTU) besides the default value in your migration network, you must also input the same value in the {ocp-short} transfer network that you use. For more information about the {ocp-short} transfer network, see xref:creating-migration-plan-2-8_rhv[Creating a migration plan].
====
```

### PREVIEW

* [6.1. Adding a VMware vSphere source provider](https://anarnold97.github.io/MTA-Preview/MTV/snippet-fix.html#adding-source-provider_vmware)


RichardHoch:
@anarnold97  LGTM!  

The snippet was introduced in https://github.com/kubev2v/forklift-documentation/pull/578, which has other changes that need to be added -- and some that are now irrelevant. I'll specify the relevant changes in the list I prepare.

PR #660, opened 2025-03-21
MTV-2228: Remove modules that are no longer needed

anarnold97:


documentation/modules/creating-network-mapping.adoc
RichardHoch, line 1: Please leave in repo -- will save us time when this UI is fixed, even if it is changed as I expect.

documentation/modules/creating-storage-mapping.adoc
RichardHoch, line 1: Please leave in repo -- will save us time when this UI is fixed, even if it is changed as I expect.

RichardHoch:
Two files should be retained (see Files changed); otherwise LGTM.

PR #674, opened 2025-04-07
Maps left in UI after archiving and deletion

RichardHoch:
MTV 2.8.1

Resolves https://issues.redhat.com/browse/MTV-2172 by adding a note to the "Plan options" section of the MTV user guide.

Preview: https://file.corp.redhat.com/rhoch/maps_left_in_ui_after_deletion/html-single/#migration-plan-options-ui_vmware [final note; same for all providers]

### DIFF
--- a/documentation/modules/migration-plan-options-ui.adoc
+++ b/documentation/modules/migration-plan-options-ui.adoc
@@ -42,3 +42,12 @@ On the *Plans for virtualization* page of the {ocp} web console, you can click t
 
 Deleting a migration plan does not remove temporary resources. To remove temporary resources, archive the plan first before deleting it.
 ====
++
+[NOTE]
+====
+The results of archiving and then deleting a migration plan vary by whether you created the plan and its storage and network mappings using the CLI or the UI.
+
+* If you created them using the UI, then the migration plan and its mappings no longer appear in the UI.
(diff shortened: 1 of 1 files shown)


documentation/modules/migration-plan-options-ui.adoc
```diff
@@ -42,3 +42,9 @@ On the *Plans for virtualization* page of the {ocp} web console, you can click t
 
 Deleting a migration plan does not remove temporary resources. To remove temporary resources, archive the plan first before deleting it.
 ====
++
+[NOTE]
+====
+If you archive and then delete a plan that was created by using the CLI, its Network Mapping and Storage Mapping information might remain visible in the UI.
```
solenoci, line 48: This is accurate, however I think it would be great if we mentioned the difference between creating the plan and rest of the resources through UI vs. through CLI.

UI: All resources will be deleted alongside plan.

CLI: Resources (maps) may stay on the cluster for further use by different plans (intended behavior). Bugfix for MTV-2172 is that the migration resource will be deleted even when created using CLI.

RichardHoch:
@solenoci I rewrote the note based on your comment. Please review the PR again. Thanks.

solenoci (approved):

RichardHoch:
@anarnold97 Please review this PR. If good, please merge. OK to publish as well.

anarnold97 (approved):

PR #679, opened 2025-04-15
Option to not map network during creation of CLI migration plan

RichardHoch:
MTV 2.8.1

Resolves https://issues.redhat.com/browse/MTV-2357 by adding a new valid value, `ignored`, to `spec:map:destination:type` in the `NetworkMap` CR. 

Preview: https://file.corp.redhat.com/rhoch/no_map_network_cli/html-single/#new-migrating-virtual-machines-cli_vmware [step 4, callout 1]

### DIFF
--- a/documentation/modules/new-migrating-virtual-machines-cli.adoc
+++ b/documentation/modules/new-migrating-virtual-machines-cli.adoc
@@ -396,7 +396,7 @@ spec:
       namespace: <namespace>
 EOF
 ----
-<1> Allowed values are `pod` and `multus`.
+<1> Allowed values are `pod`, `multus`, and `ignored`. Use `ignored` to avoid attaching VMs to this network for this migration. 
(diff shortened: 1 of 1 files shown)


documentation/modules/new-migrating-virtual-machines-cli.adoc
```diff
@@ -396,7 +396,7 @@ spec:
       namespace: <namespace>
 EOF
 ----
-<1> Allowed values are `pod` and `multus`.
+<1> Allowed values are `pod`, `multus`, and `ignored`. Use `ignored` to avoid migrating VMs to this network for this migration.
```
Pkylas007, line 399: Suggestion: Use `ignored` to avoid migrating this network for target VMs.

RichardHoch: @Pkylas007 Thanks -- I thought of that previously, but it could be understood to mean in all cases. Here I really want to specify that this is a one-time only situation, which, of course, be repeated. Of course, @mnecas will correct me if I'm wrong.

RichardHoch: @Pkylas007 @mnecas And just like that, he corrected me! :)

mnecas (approved):
lgtm

anarnold97 (approved):

PR #690, opened 2025-04-30
HOLD FOR MTV 2.8.3: Add shared parameter to PVC name template

RichardHoch:
MTV 2.8.3

Resolves https://issues.redhat.com/browse/MTV-2339 by adding the `shared` parameter to the PVC name template.

Preview: https://file.corp.redhat.com/rhoch/pvc_shared/html-single/#new-migrating-virtual-machines-cli_vmware [step 8, callout 10. Note that callout 16 refers to callout 10, so callout 16 does not need to be changed.]

### DIFF
--- a/documentation/modules/new-migrating-virtual-machines-cli.adoc
+++ b/documentation/modules/new-migrating-virtual-machines-cli.adoc
@@ -745,7 +745,7 @@ spec:
     LS0tCi0gbm... <2>
 EOF
 ----
-<1> (Optional) {ocp} service account. The `serviceAccount` must be provided if you want to manipulate any resources of the cluster.
+<1> Optional: {ocp} service account. Use the `serviceAccount` parameter to modify any cluster resources.
 <2> Base64-encoded Ansible Playbook. If you specify a playbook, the `image` must include an `ansible-runner`.
 +
 [NOTE]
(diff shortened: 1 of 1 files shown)


documentation/modules/new-migrating-virtual-machines-cli.adoc
```diff
@@ -847,10 +847,12 @@ The template follows the Go template syntax and has access to the following vari
 * `.PlanName`: Name of the migration plan.
 * `.DiskIndex`: Initial volume index of the disk.
 * `.RootDiskIndex`: Index of the root disk.
+* `.Shared`: `true`, if the volume is shared by multiple VMs, `false`, if it is not.
```
shdeshpa07, line 850:
```suggestion
* `.Shared`: `true`, if multiple VMs share the volume. 
```
I am assuming that `false` is implied here. Feel free to bring it back, if you think otherwise.

shdeshpa07 (commented):
LGTM. Just a few minor nits.

documentation/modules/new-migrating-virtual-machines-cli.adoc
RichardHoch: @shdeshpa07 Whether or not `false` is implied, it's a valid option, so it needs to be documented, IMHO.

RichardHoch: @shdeshpa07 Take a second look, please?

shdeshpa07 (commented):
LGTM.

RichardHoch:
@mnecas Please review this PR. Thanks.

mnecas (approved):
lgtm

PR #757, opened 2025-09-04
MTV-2765 remove inclusive module release notes

Jenny-Anne:
https://issues.redhat.com/browse/MTV-2765


anarnold97 (changes requested):
LGTM - but if we are removing the module completely, should we delete the module?
WDYT

anarnold97 (approved):

